PR #44, opened 2017-06-19
Cif2 conversion: removed aliases that match the _definition.id

vaitkus:


vaitkus:
Thank you for the merge.

PR #84, opened 2018-06-28
Fix example of _atom_sites_Cartn_transform_axes

pavoljuhas:
When `a` is parallel to `x`, `b` needs to be in the `xy` plane.
The `b in yz` orientation is impossible when gamma != 90deg.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -20596,7 +20596,7 @@ _type.container                         Single
 _type.contents                          Text
 loop_
   _description_example.case
-         'a parallel to x; b in the plane of y & z' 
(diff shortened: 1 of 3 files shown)


jamesrhester:
This exposes an issue with the definition of the Cartesian transform matrix, which as the definitions are written is free choice. I have raised issue #85 to double-check that the Rollett convention is the only one needed.

jamesrhester:
Sorry for the long delay on this, I needed to make sure we could accommodate alternative Cartesian axis specifications in the current dREL framework: answer, yes we can.

Issue #31, opened 2017-04-21
cif_core.dic (cif2-conversion): _type.dimension notation for matrices


vaitkus:
The ``DDLm`` definition of the ``_type.dimension`` data item provides the following  example for matrices:
```
loop_
_description_example.case
_description_example.detail
"[3,3]"              '3x3 matrix of elements'  
```
however in the ``cif_core.dic`` (and ``cif_ms.dic``) contain multiple entries where the elements in brackets are separated by whites paces instead of a comma, i. e. the ``_cell.metric_tensor`` data item
contains the following value:
```
_type.dimension                         '[3  3]'
```

Which case should be treated as the correct one?

jamesrhester:
The comma notation conforms to the original DDLm paper.  CIF2 uses spaces to separate list elements, and so if _type.dimension were not quoted whitespace would be appropriate. I think that commas should be the delimiter, but if you have a strong reason for preferring whitespace it might be possible to make this change, as almost no software that I know of actually uses this information.

vaitkus:
I do not have any objections to using commas as separators. I just noticed the inconsistency between the example and the actual usage.

However, I would like a more formal description of the internal structure of the '_type.dimension' string for parsing/validation purposes. Should the string be treated as potentially having the format described in #18 or is it a bit simpler now? If needed, I could translate the human-readable definition into a more formal one (something similar to what I proposed in #34).

jamesrhester:
I would prefer to wait before spending time on defining a more intricate format. If we can express all current dictionaries without using a more complex structural definition, that would be good as it would save programmer time.  Ideally it would just become a simple list of integers.

vaitkus:
So do we agree to consistently use commas as separators in **_type.dimension** values?

If yes, I can created a pull request that would homogenise the notation across existing DDLm dictionaries.

jamesrhester:
Sorry, I missed this question last year(!). Yes we can agree to use commas as separators.

jamesrhester:
Yes, that is correct. The original DDLm project envisioned nested dimensions, but the need for this appears to be an artefact of the way in which artificial keys were constructed.

vaitkus:
The description of the 'Dimension' data type has been slightly updated and the agreed upon notation propagated across the rest of the existing dictionaries. The changes appear in PR #121

vaitkus:
Closing the issue since it has been resolved.

PR #111, opened 2019-01-07
Cif2 conversion type dimension fix

vaitkus:
Adding mandatory quotes around the values of the type dimension data item.
Adding commas instead of spaces as separators as discussed in issue #31.


jamesrhester:
Thanks.

PR #133, opened 2019-06-04
Assigned more descriptive _type.contents values to several id data items.

vaitkus:
Several id data items (``_geom_angle.id``, ``_geom_bond.id``, etc.) were assigned a ``List`` content type.
This PR replaces them with data type lists (i.e. ``List(Code,Symop)``, ``Code,Symop``) that are compatible with the data values assigned via the dREL evaluation methods.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
 _dictionary.title                       CORE_DIC
 _dictionary.class                       Instance
 _dictionary.version                     3.0.11
-_dictionary.date                        2019-04-03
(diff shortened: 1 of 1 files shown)


jamesrhester:
I think it would be better to simply remove these synthetic key data names. They were created as synthetic identifiers before there was a way to reference the values of multiple key data names when indexing into a category in dREL using the [] notation.  They are essentially a dREL implementation detail.

vaitkus:
Closing the PR since the proposed changes are no longer relevant. See issue #149 for a more in-depth discussion.

PR #137, opened 2019-06-25
Using a resolving URI for cif_core.dic

merkys:
`cif_core.dic` has `_dictionary.uri` of `www.iucr.org/cif/dic/cif_core.dic`, which lacks a scheme prefix (i.e. `http://`) to conform to [RFC 3986](https://tools.ietf.org/html/rfc3986). Moreover, this URI does not resolve. Thus I suggest using GitHub URI (`https://raw.githubusercontent.com/COMCIFS/cif_core/cif2-conversion/cif_core.dic`) as `_dictionary.uri` for this dictionary.


merkys:
Thanks, @jamesrhester!

PR #152, opened 2019-09-25
Added the DICTIONARY_AUDIT loop in the 'cif_img.dic' dictionary

vaitkus:
The loop is needed to properly document additional changes to the dictionary (i.e. the removal of the "Count" data type).

The audit history compiled so far was based of GitHub logs and as such may be erroneous/incomplete. Please review and change accordingly if needed.

### DIFF
--- a/cif_img.dic
+++ b/cif_img.dic
@@ -14,7 +14,7 @@ data_CIF_IMG
 
     _dictionary.title            CIF_IMG
     _dictionary.class            Instance
-    _dictionary.version          3.00.02
+    _dictionary.version          3.00.03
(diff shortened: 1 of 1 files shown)


jamesrhester:
Note that the imgCIF dictionary has a separate repository in which it is still in DDL2. This repository should become the focus of changes.  However, we can continue to improve the DDLm version and merge in future.

vaitkus:
Acknowledged. I will only propose changes in this repo for this dictionary that are related to the CIF2/DDLm aspects of the file rather than the semantics.

Issue #150, opened 2019-09-25
Change "count" and "index" types to be "integer"


jamesrhester:
Following #140 , all instances of "Count" or "Index" types in official DDLm dictionaries should be changed to "Integer", with ``_enumeration.range`` set appropriately, following which these should be marked as deprecated in the ``type.contents`` definition.

jamesrhester:
I have just merged cif2-conversion with master, and would like us now to switch to using master as the default branch.

vaitkus:
Great news. I will issue all of my minor pull request directly to master in the future.

However, since the latest version of the ``cif2-conversion`` branch did contain several additional commits (from the #152, #154 PRs), changes related to this issue were issued as a PR (#157)  to the branch rather than the trunk. If this is not the preferred way, please close the pull request and I will raise a new one directly to ``master``.

jamesrhester:
This work has now been merged. Thanks for the great janitorial work.

PR #157, opened 2019-09-26
Deprecated 'Count' and 'Index' content types

vaitkus:
This PR implements changes outlined in issue #150.

The types were marked as deprecated in the ``type.contents`` definitions by simply adding a '(deprecated)' string to their human-readable descriptions. Please modify accordingly if this is insufficient.


jamesrhester:
Thanks for the updates.

PR #166, opened 2019-12-09
Fixed: corrected the usage of the kelvin measurement units 

vaitkus:
Kelvins are not measured in degrees.


jamesrhester:
Thanks for the improvement.

PR #190, opened 2021-01-06
Expand the list of official _audit.schema

jamesrhester:
The dictionaries supported by the IUCr adopt a variety of schemas. These proposed values for `_audit.schema` document them for use in data files conforming to the particular dictionaries.


jamesrhester:
I am leaving this open as there may be discussion soon on an additional entry.

jamesrhester:
I think this is now outdated with the updated multiple-block approach. It is likely we can get by with simply `Default` and `Custom`.

PR #192, opened 2021-02-02
Clarification of Measurand - SU data item relationship.

jamesrhester:
As discussed in ddlm-group, clarify the relationship between SU data names and measurand data names and how they may be presented in a data file.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -8,11 +8,11 @@ data_DDL_DIC
 
     _dictionary.title            DDL_DIC
     _dictionary.class            Reference
-    _dictionary.version          4.0.0
-    _dictionary.date             2020-10-30
+    _dictionary.version          4.0.1
+    _dictionary.date             2021-02-02
     _dictionary.uri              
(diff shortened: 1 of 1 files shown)


vaitkus:
During the same discussion in the ddlm-group [1] it was also agreed that the phrase "must be provided" could be changed to "should be provided". However, this is currently not reflected in the PR.

[1] https://www.iucr.org/__data/iucr/lists/ddlm-group/msg01688.html

jamesrhester:
I may have been misunderstood: the phrase "should be provided" was intended to refer to the appearance of the su value in a _data file_. I think that a _definition_ for the su value _must_ be provided in the dictionary, otherwise there is nothing for a su value appearing in a data file to be associated with.

vaitkus:
Sorry. that is a mistake on my part. The proposed definition is clear, I simply misread it. The "*must*" discussed in the ddlm-group actually was properly changed to "*should*".

Issue #194, opened 2021-02-17
Suggestion to remove the default value of the _units.code data item


vaitkus:
The DDLm reference dictionary specifies that the `_units.code` data item has the default value of "`arbitrary`". This raises several problems:

1. Data items with non-quantifiable values (i.e. text strings) are automatically assigned measurement units although they do not really apply;
2. Data items with quantifiable values, but without explicit measurement units are assigned the "`arbitrary`" units although it is just as likely (or even more likely) that the values are simply dimensionless (marked as "`none`" according to the DDLm reference dictionary, "one" or "1" according to SI [1]).

       
For example, none of the data items in the `CIF_CORE` dictionary with the Integer data type have explicit measurement units. Due to this, all of them default to the "arbitrary" measurement units, although most of them should most likely be marked as dimensionless (i.e. `_exptl.crystals_number`, `_reflns_shell.number_unique_gt`).

Based on this, I propose to completely remove the default value for the `_units.code` data item. This would solve both problems since incorrect assumptions would be avoided for both the quantifiable and  non-quantifiable items. Of course, one would still have to assign explicit `_units.code` to the aforementioned Integer items, however, that seems to fit well with the recommended approach.

Also, if non-quantifiable items were also required to have an explicit `_units.code` value, I would suggest using the non-applicable value ('.') to denote this as opposed to the 'none' enumeration value.

[1] https://en.wikipedia.org/wiki/Dimensionless_quantity

jamesrhester:
I agree with these suggestions, and I haven't seen any objections. Could you create a pull request that removes the default value and adds 'none' to the dimensionless quantities?

vaitkus:
Sure. PR #197 should solve the outlined issue.

vaitkus:
I have discovered that I made some errors in my previous PR. I have fixed them in PR #204, but I would like to keep the issue open until the new PR is merged.

PR #197, opened 2021-03-01
Remove default units code

vaitkus:
This PR solves issue #194 by removing the default value of the `_units.code` data item ('`arbitrary`') and adding the `none` units to the definitions of multiple dimensionless data items. The definitions of several `_refine_diff.density_*_su` data items were also modified by copying the dREL methods for measurement units resolution from the main items that they are linked to (`_refine_diff.density_*`).


jamesrhester:
Thanks for the careful work.

PR #200, opened 2021-03-01
Fix misspelt data names

vaitkus:
This PR fixes a few misspelt data names in the newly added data item definitions.


jamesrhester:
Thanks for this fix. I have removed mention of the correction to the attributes as it would be irrelevant to anyone reading the history. Hopefully automatic dictionary checking will catch these mistakes in the future.

Issue #203, opened 2021-03-18
Unit of measurement of the _model_site.adp_eigen_system data item


vaitkus:
The `_model_site.adp_eigen_system` is defined as a matrix that contains both the eigenvalues and the corresponding eigenvectors. The following evaluation method in provided in the definition:
```
      A    =  _cell.orthogonal_matrix
      U    =  A * _model_site.adp_matrix_beta * Transpose(A) /(2*Pi**2)
 
      _model_site.adp_eigen_system  =  Eigen( U )
```
where `_cell.orthogonal_matrix` is measured in A (`angstroms`) and  `_model_site.adp_matrix_beta` is dimensionless (`none`). From this it would seem matrix U is measured in A^2 (`angstrom_squared`) and so should the eigenvalues **or** the eigenvectors of U as the multiplication of these should result in an object with the same unit of measure as U.

However, since the Eigen(U) contains both the eigenvalues **and** the eigenvectors it is not clear what units of measure should be assigned (`angstrom_squared`, `none` or some kind of  an element-wise mixture of both).

jamesrhester:
Yes, this is a problem with our system. Ideally we should split this into two data names. I will raise it with the DDLM-group.

jamesrhester:
The ddlm group has no objections. I will prepare a pull request.

vaitkus:
PR #214 was successfully merged.

PR #204, opened 2021-03-18
Fix units of measure

vaitkus:
This PR fixes several issues related to units of measure that were accidentally introduced in PR #197. The main changes:

- several definitions were updated by adding the unit of measure;
- several definitions were updated by changing the unit of measure from 'none' to the more appropriate one.

I tried to be more attentive with the units this time, but it would be great if someone with a keen eye could double check the changes.

Also, there seems to be some ambiguity with the units of measure of the `_model_site.adp_eigen_system` data item (see issue #203). I have changed the units of measure from `none` to `angstrom_squared`, however, I am not completely sure if any of them are actually correct for all elements of the matrix.


jamesrhester:
The issue with `_model_site.adp_eigen_system` is an important one. This is, as far as I can tell, the only matrix-valued item for which units could be different for different elements, and our units formalism only allows a single unit to be expressed. It only exists for computational convenience, as getting eigenvalues and eigenvectors at the same time is efficient.  I believe that the ideal solution would be to deprecate this and replace it with two separate items corresponding to the eigenvectors and eigenvalues. I will raise an issue.

PR #209, opened 2021-06-24
Add automated check of DDLm conformance after fixing minor errors

jamesrhester:
A DDLm check action has been added. A number of minor changes were required to the core files to bring them into conformance:

- data names should not be aliased to themselves
- A data name should be `_type.purpose` 'Link' if it refers to a parent data name

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
 _dictionary.title                       CORE_DIC
 _dictionary.class                       Instance
 _dictionary.version                     3.0.14
-_dictionary.date                        2021-03-03
+_dictionary.date                        2021-06-29
 _dictionary.uri
 https://raw.githubusercontent.com/COMCIFS/cif_core/cif2-conversion/cif_core.dic
 _dictionary.ddl_conformance             4.0.1
@@ -484,7 +484,7 @@ save_DIFFRN_ATTENUATOR
 _definition.id                          DIFFRN_ATTENUATOR
 _definition.scope                       Category
 _definition.class                       Loop
-_definition.update                      2013-09-08
+_definition.update                      2021-06-29
 _description.text                       
 ;
      The CATEGORY of data items which specify the attenuators used in the
@@ -1138,7 +1138,7 @@ save_DIFFRN_ORIENT_REFLN
 _definition.id                          DIFFRN_ORIENT_REFLN
 _definition.scope                       Category
 _definition.class                       Loop
-_definition.update                      2013-09-08
+_definition.update                      2021-06-29
 _description.text                       
 ;
      The CATEGORY of data items which specify the reflections used to
@@ -1563,7 +1563,7 @@ save_DIFFRN_RADIATION_WAVELENGTH
 _definition.id                          DIFFRN_RADIATION_WAVELENGTH
 _definition.scope                       Category
 _definition.class                       Loop
-_definition.update                      2013-09-08
+_definition.update                      2021-06-29
 _description.text                       
 ;
      The CATEGORY of data items which specify the wavelength of the
@@ -1734,7 +1734,7 @@ save_DIFFRN_REFLN
 _definition.id                          DIFFRN_REFLN
 _definition.scope                       Category
 _definition.class                       Loop
-_definition.update                      2013-09-08
+_definition.update                      2021-06-29
 _description.text                       
 ;
      The CATEGORY of data items which specify the reflection measurements,
@@ -2864,7 +2864,7 @@ save_DIFFRN_REFLNS_CLASS
 _definition.id                          DIFFRN_REFLNS_CLASS
 _definition.scope                       Category
 _definition.class                       Loop
-_definition.update                      2013-09-08
+_definition.update                      2021-06-29
 _description.text                       
 ;
      The CATEGORY of data items which specify different classes of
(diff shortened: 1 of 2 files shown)


vaitkus:
It seems that I slightly too hasty while merging a recent PR in the ddlm check action repo (see [1]) and overlooked an error that causes the dictionary validation against the ddlm dictionary to always successfully pass regardless of the actual errors . Upon fixing this issue, the ddlm action should pick up some additional issues in the cif_core.dic dictionary (i.e. it still has the `_category.key_id` data item that is no longer defined in the DDLm reference dictionary).

I also noticed that the `_atom_site.label` alias `_atom_site.id` was removed even though it does not violate the alias naming policy and was seemingly at some point introduced for compatibility purposes with PDBx/mmCIF (see [2] ). I am not sure if this compatibility is still relevant, though.

Finally, I would also like to make it clear that the stylistic rule of not having aliases that match the data name is not currently specified in any official IUCr documentation or stylistic guide. The check was one-sidedly introduced by me into the cif_ddlm_dic_check program because I found the duplicate aliases superfluous. I am happy if others agree with my opinion, but I simply wanted to note the fact that it is not official IUCr policy.

[1] https://github.com/COMCIFS/dictionary_check_action/pull/5
[2] https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.id.html

cif_core.dic
```diff
@@ -19730,8 +19712,7 @@ save_atom_site.label
 _definition.id                          '_atom_site.label'
 loop_
   _alias.definition_id
-         '_atom_site_label'  
-         '_atom_site.id'
```
vaitkus, line 19734: Is it really OK to remove the _atom_site.id alias? It seems that at some point it was introduced for compatibility purposes with the PDBx/mmCIF dictionary (see [1] ).

[1] https://mmcif.wwpdb.org/dictionaries/mmcif_pdbx_v50.dic/Items/_atom_site.id.html

jamesrhester:
You are right, `_atom_site.id` should be left in the dictionary. I will update the PR.

cif_core.dic
jamesrhester: Yes, see my comment below (not sure about the difference between a review comment and a general comment on an issue?)

vaitkus:
Great, thank you for the update. However, I am a little bit worried about the outright removal of the `_category.key_id` data item. In cases where the id key matches the `_category_key.name` data item value the removal is indeed straightforward. However, in several categories the opaque key is left replaced by a set of data items (i.e. `REFLN` category had the `_refln.hkl` id key and is now left with  `_refln.index_h`, `_refln.index_k`, `_refln.index_l` key names). This might be the desired result, but in several other categories the opposite is done -- a non-opaque key is replaced by an opaque one (i.e. see the `PUBL_AUTHOR` category). It is probably not necessary to homogenise the key usage in this PR, however, it might be slightly easier to do this now than after the complete removal of the keys.

jamesrhester:
So I have reviewed the `_category.key_id` items. Cases:
1.  A single data name in `_category_key.name`  identical to `_category.key_id`: no issues
2.  points to a synthetic key that is a CIF2 list constructed from the components: 
  - if `hkl`, that item can be retained as it is useful in its own right for matrix multiplication and so on.
  - otherwise, the whole definition can be removed
3. points to an opaque key that is not in `_category_key.name`: only `chemical_conn_bond.id`, which was originally also a list. I think this can be removed as it is not referred to anywhere else in the dictionary or mmCIF.

So given the above I propose removing or deprecating `_reflns_shell.d_res_limits`, `_chemical_conn_bond.id`, `geom_*.id`, `_citation_author.key` and `_citation_editor.id`. I believe none of these will ever have appeared in any data files, but for good discipline I suppose the dictionary major version should be bumped if these definitions are deleted instead of deprecated.

jamesrhester:
Also, my draft layout check appears to have made it into the workflow for COMCIFS/cif_core, I thought it was restricted to my personal repository. Please don't worry about that check failing at this point, we will do a separate commit once we are happy with the style guide and results. We may even use a different layout checker if e.g. COD produces one.

vaitkus:
Ok, this clears up things quite a bit. I do not mind those items existing, deprecated or not, since they may generally be useful in dREL operations. You are clearly more qualified to judge the utility of such items. My question is slightly different -- should the `_category_key.name` data item refer to the opaque identifiers, combined identifiers (i.e. lists) or to the items that form natural keys? For example, key of `DIFFRN_REFLN` is `_diffrn_refln.hkl` (list), key of `DIFFRN_ORIENT_REFLN` is the combination of `_diffrn_orient_refln.index_h`, `_diffrn_orient_refln.index_k`, `_diffrn_orient_refln.index_l` (natural key, could potentially be a list) while `PUBL_AUTHOR` key `_publ_author.id` (opaque, and seemingly for a good reason since author names are not guaranteed to be unique). This might seem nit-picky, but it since the category key introduces a uniqueness constraint this actually affects the validation results (i.e. list type values are often constructed using dREL statements and thus are quite more difficult to validate).

jamesrhester:
`_category_key.name` should always specify the data names that we have decided should be the category key, whether they are synthetic (as has been the case in the past) or opaque. `_diffn_refln.hkl` is then purely a convenience data name to represent the vector hkl for use in dREL methods. That it is also able to act as a key does not need to be captured, but could be deduced.

In the future I would like to stop indexing reflection packets with h,k,l as these values can be repeated. But that's for the future.

I've realised my github/git/branching mistake that resulting in extra stuff being committed, but it looks like it is just the action-related items so I think it is OK to leave them in.

vaitkus:
Great, thank you for the explanation. Note, that I made a few minor tweaks before merging the PR (DDL conformance number, definition update dates, etc.).

PR #214, opened 2021-07-01
Replaced model_site.adp_eigen_system

jamesrhester:
This pull request replaces `_model_site.adp_eigen_system` with separate eigenvalue and eigenvector data names. I have opted to remove eigen_system completely, even though that would normally be a violation of standard rules, because eigen_system is nowhere used in the wild. 

Please comment on whether it would be more appropriate instead to use the `definition_replaced.by` attributes and not remove the eigen_system data name.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -13398,7 +13398,7 @@ _type.purpose                           Measurand
 _type.source                            Derived
 _type.container                         Array
 _type.contents                          Real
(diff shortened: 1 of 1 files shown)


vaitkus:
I merged the PR since no objections to the removal of the `_model_site.adp_eigen_system` data item were voiced.

PR #217, opened 2021-07-07
Updated the _journal_index.type data item definition

vaitkus:
This PR updates the definition of the `_journal_index.type` data item to match the description in ITC Volume G (2005), section 3.2.5.4.

I would have also adapted the 3.2.5.2 example to the `JOURNAL_INDEX` category definition, however, I was unsure if this is allowed (the current definition of the `_description_example.case` data item seems to imply that this item should only be used in data item definitions).

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
 _dictionary.title                       CORE_DIC
 _dictionary.class                       Instance
 _dictionary.version                     3.1.0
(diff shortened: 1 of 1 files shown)


jamesrhester:
`_description_example.case` can be used in both category and item definitions. If the current definition seems to imply otherwise, perhaps an improvement in the text is worth considering? Or maybe explicitly adding the description category to `dictionary_valid`?

Issue #220, opened 2021-07-15
ddl.dic: it is not clear if items from the DESCRIPTION_EXAMPLE category can be used in category definitions


vaitkus:
It was noted in the discussion of PR #217 that it is not clear from the dictionary definition that the `_description_example.case` data item can be used in category definitions. It was suggested [1] that:

1. The data item definition could be improved to better reflect the possible usage.
2. The possible usage could be expressed using the `DICTIONARY_VALID` loop.

### Updating the definitions
In my opinion, the definitions of the `DESCRIPTION_EXAMPLE` category as well as the `_description_example.case` and `_description_example.detail` data items currently imply that the category only applies to data item definitions, i.e.:
```
The attributes of descriptive (non-machine parsable) examples of values of the defined items.
```
The descriptions could most likely be easily extended to include categories, however, special attention should be given to the `_description_example.case` data item. Currently, it states that:
```
An example case of the defined item. Instances of this data item inherit the container,
content and purpose type constraints of the defining item.
```
However, since category definitions are prohibited from containing attributes from the `TYPE` category, the container, content and purpose attributes cannot be inherited from a category definition. One may potentially rely on the default values of these attributes ('Single', 'Text', 'Describe'), however, it might be better to explicitly state this as the default values may change in the future.

### Updating the `DICTIONARY_VALID` loop
The `DICTIONARY_VALID` loop allows to specify which attributes are mandatory, recommended or prohibited in certain contexts (Dictionary, Category, Item). However, currently, there is no direct way to specify that the item is *allowed* in a certain context. One may rely on the item not being explicitly prohibited, but I am unsure how reliable this is. Updating the descriptions seems like a better solution.

[1] https://github.com/COMCIFS/cif_core/pull/217#issuecomment-876872709

jamesrhester:
#223 addresses the second point. The wording of the `description_example` category should be improved.

jamesrhester:
#224 should address the remainder of this issue.

vaitkus:
Great, thank you.

PR #223, opened 2021-07-20
Clarify status of attributes not mentioned in dictionary_valid loop.

jamesrhester:
Addresses part of issue #220


jamesrhester:
Note based off master before pull request #222, thus the layout check failure. Probably best to merge this first?

PR #227, opened 2021-07-21
Clarify meaning of _dictionary_valid.attributes

jamesrhester:
Make sure it is clear that child categories are included when a category is given in the list of attributes in the dictionary_valid list.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -772,11 +772,12 @@ save_dictionary_valid.attributes
 
     _definition.id                '_dictionary_valid.attributes'
     _definition.class             Attribute
-    _definition.update            2013-03-06
+    _definition.update            2021-07-21
     _description.text
 ;
     A list of the attribute names and categories that are assessed
-    for application in the item, category and dictionary definitions.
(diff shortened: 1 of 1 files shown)


ddl.dic
```diff
@@ -772,11 +772,12 @@ save_dictionary_valid.attributes
...
     _definition.id                '_dictionary_valid.attributes'
     _definition.class             Attribute
-    _definition.update            2013-03-06
+    _definition.update            2021-07-21
     _description.text
 ;
     A list of the attribute names and categories that are assessed
-    for application in the item, category and dictionary definitions.
+    for application in the item, category and dictionary definitions. A
+    parent attribute category implicitly includes all child categories.
```
vaitkus, line 780: I propose that the definition be further improved by explicitly stating that the inclusion of child categories is recursive, that is, that it also includes the (grand)+child categories. I.e. `A parent attribute category implicitly includes all child categories in a recursive manner.` or something similar, but better sounding.

vaitkus: Great, thank you for the revision.

Issue #207, opened 2021-03-19
templ_attr.cif: units of measurement in the _fract_matrix save frame


vaitkus:
The units of measurement provided in the `_fract_matrix` save frame of the `templ_attr.cif`  file seem incorrect.

The `_fract_matrix` save frame describes elements of the matrix and vector that are used to transform Cartesian coordinates (in angstrom) to fractional coordinates (dimensionless) as described in the definition:
```
       x'                  |11 12 13|     x                  | 1 |
     ( y' )fractional = mat|21 22 23| * ( y )Cartesian +  vec| 2 |
       z'                  |31 32 33|     z                  | 3 |
```

The save frame lists `none` as the units, but this seems to only apply to the vector elements while the matrix elements should have the units of `reciprocal_angstroms`.

One way to resolve this would be to refactor the current save frame into two save frames (`_fract_matrix` and `_fract_vector`) and set the units accordingly to `reciprocal_angstroms` and `none`. Does this seems reasonable?

Also, for the sake of consistency, the same refactoring could be done to the `_Cartn_matrix` save frame that describes elements of the matrix and vector responsible for the reverse transformation (fractional to Cartesian), even though  both the matrix and the vector have the same units (see #206). Does this additional change also seems reasonable?

jamesrhester:
Yes, this should be fixed, I think it was a cut-paste oversight.

vaitkus:
Issue was resolved by merging PR #230.

Issue #206, opened 2021-03-19
templ_attr.cif: units of measurement in the _Cartn_matrix save frame


vaitkus:
The units of measurement provided in the `_Cartn_matrix` save frame of the `templ_attr.cif` file seem incorrect.

The `_Cartn_matrix` save frame describes elements of the matrix and vector that are used to transform fractional coordinates (dimensionless) to Cartesian coordinates (in angstrom) as described in the definition:
```
       x'                  |11 12 13|     x                  | 1 |
     ( y' )Cartesian  = mat|21 22 23| * ( y )fractional + vec| 2 |
       z'                  |31 32 33|     z                  | 3 |
```
The save frame lists `reciprocal_angstroms` as the units, but shouldn't both the matrix and the vector have `angstroms` as the units?

jamesrhester:
Yes, these should be `angstroms`.

vaitkus:
Issue was resolved by merging PR #230.

PR #230, opened 2021-07-21
Fix coordinate matrix unit of measurement

vaitkus:
This PR addresses issues #206 and #207.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -20648,7 +20648,7 @@ save_
 save_atom_sites_fract_transform.matrix
 
     _definition.id                '_atom_sites_fract_transform.matrix'
-    _definition.update            2021-03-03
+    _definition.update            2021-07-21
     _description.text
 ;
     Matrix used to transform Cartesian coordinates in the ATOM_SITE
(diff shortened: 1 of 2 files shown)


jamesrhester:
I think we can add `_type.dimension` to the new template definitions now that matrix and vector have been split. I'm not sure how to automatically add this to your pull request using Github.

vaitkus:
> I think we can add `_type.dimension` to the new template definitions now that matrix and vector have been split.
This might not be clear from the template description text, but the template save frames are not imported by a matrix or vector definition, but rather by definitions of the individual matrix or vector elements (i.e. `_atom_sites_Cartn_transform.vec_1`, `atom_sites_Cartn_transform.vec_2`, etc.). Therefore, the `_type.dimension` probably does not apply here. Please correct me if I misunderstood your suggestion.

Also, the template descriptions could probably be rewritten in a more clear manner that shows that they apply to individual matrix/vector elements, however, I am unsure if this is in the scope of this PR. Nevertheless, suggestions are welcome.

> I'm not sure how to automatically add this to your pull request using Github.
Usually, when only minor changes are needed I tend to modify the PR files directly through the GitHub web interface. Alternatively, I think that cloning my repo and simply checking out the PR branch should do the trick as I always allow the maintainers of the project to edit my PRs. For more information see [1].

[1] https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/committing-changes-to-a-pull-request-branch-created-from-a-fork

Issue #245, opened 2021-09-13
Items have become case-insensitive in DDLm core relative to DDL1


jamesrhester:
Both `_atom_site.label` and `_space_group_symop.id` are case-sensitive in DDL1 but have type `Code` in the DDLm dictionary, which means they are case-insensitive. I propose reverting these and any other such changes to be case-sensitive.

PR #249, opened 2021-09-16
Change caseless data values to cased for consistency with DDL1.

jamesrhester:
In the transition to DDLm it appears that some data names were
classed as 'Code' type (caseless) when in DDL1 they were 'char'
(always case-sensitive). This patch attempts to identify those
data names for which this difference is significant. In some
situations, such as data names referring to external
database identifiers, the identifiers should be caseless, even
if there was no mechanism in DDL1 to indicate this.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -491,7 +491,7 @@ save_diffrn_attenuator.code
     _type.purpose                 Encode
     _type.source                  Assigned
     _type.container               Single
-    _type.contents                Code
+    _type.contents                Text
 
 save_
 
@@ -1617,7 +1617,7 @@ save_diffrn_radiation_wavelength.id
     _type.purpose                 Encode
     _type.source                  Assigned
     _type.container               Single
-    _type.contents                Code
+    _type.contents                Text
     _description_example.case     x2
 
 save_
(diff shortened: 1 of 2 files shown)


jamesrhester:
This pull request is aimed at fixing #245. I have also attempted to catch any other items that were case-sensitive in DDL1 but not in DDLm. I have assumed that references to external identifiers (e.g. databases) are case-insensitive.

vaitkus:
Firstly, I wanted to note, that this also reintroduces ASCII whitespaces into the set of symbols that are allowed in values of the changed data items. This increases the compatibility with DDL1 even more, however, personally I somewhat liked the additional restriction of not having spaces in atom labels, etc. Not sure if this is much of an issue for other users, though.

Secondly, if I am not mistaken, one of the new desired DDLm features was that enumeration values would be case-insensitive. The current PR version would break this for `_diffrn_source.target`, `_exptl_crystal_appearance.hue` and `_exptl_crystal.colour` data items (the last one being a composition of 3 enumeration set items). Maybe these items could be left unchanged since other enumeration set items were not modified?

vaitkus:
Thank you for the changes. So just to confirm, the reintroduction of ASCII whitespaces into the set of permitted characters is not viewed as a problem?

jamesrhester:
Well it would be a problem if software authors have implicitly assumed that there would be no whitespace in the labels. We would need to invent a new `_type.contents` that was the equivalent of `Code` but case-sensitive. I'll ask on the cif-developers list to see what the feeling is.

PR #250, opened 2021-09-16
Fix definitions for b_iso and u_iso.

### LINKED ISSUE #246: Definition text for _atom_site.U_iso_or_equiv incorrect
(Raised by Alexander Marx)

it seems that the definition of U(equiv) in  `_atom_site.U_iso_or_equiv` 
is not correct as it differs from the definition
in the cited reference (Fischer & Tillmanns, 1988).

Instead of "a\~i\~ a\~j\~", the product of real-space cell lengths, the
reference contains the scalar product of the real-space unit vectors
(see also Trueblood et al., Acta Cryst A52, 770, equ. 55).

I would also suggest to add the equation for the equivalent Debye-Waller
factor to be used with U(equiv), for example:

T(equiv) = exp{-8 pi^2^ U(equiv) s^2^}, with s = sin(theta)/lambda

jamesrhester:
This pull request fixes #246

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -18193,9 +18193,9 @@ save_atom_site.b_iso_or_equiv
     atomic displacement parameter, B(equiv), in angstroms squared,
     calculated from anisotropic temperature factor parameters.
 
-        B(equiv) = (1/3) sum~i~[sum~j~(B^ij^ a*~i~ a*~j~ a~i~ a~j~)]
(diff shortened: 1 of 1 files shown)


vaitkus:
The updated formulae now match the ones in the cited reference paper. However, the correction is quite hard to detect so maybe an entry in the DICTIONARY_AUDIT would also be useful?

vaitkus:
Great, thanks.

PR #253, opened 2021-09-22
Added missing SU data names.

jamesrhester:
A recent update to DDL stipulated that all "Measurand" data names
should have SU data names defined for them. This update adds
missing SU data names for those Measurand data names that did
not previously have them.

PR #254, opened 2021-09-23
Add su 2

jamesrhester:
This is a reworked version of #253 . This time I've added a template for single-valued SU data names to save a little space, as well as including the missing Measurands, adding periods at the end of the description text, and copying the dREL methods that determine the units.


vaitkus:
Looks great, merged.

Issue #71, opened 2017-11-21
cif_core.dic: proposal to make _atom_site.Wyckoff_symbol an enumerator


vaitkus:
The ``_atom_site.Wyckoff_symbol`` is currently allows any values fitting the Encode type format. A somewhat related data item ``_space_group_Wyckoff.letter``, however, is defined as a set of enumerator values. I propose moving the enumerator values into the ``templ_attr.cif`` file and declaring both data items as importing these enumerator values.

jamesrhester:
I agree with this proposal and will implement it on a future commit.

vaitkus:
Thank You.

vaitkus:
@jamesrhester , maybe I could take a crack at this?

PR #256, opened 2021-09-23
Redefine the _atom_site.Wyckoff_symbol data item as an enumerator

vaitkus:
This PR resolves issue #71. Please note, that I also changed the data type of the `_atom_site.Wyckoff_symbol` data item from `Code` to `Text` to make it case sensitive (and match the definition of the `_space_group_Wyckoff.letter` data item). I guess for most "custom" enumeration values the distinction between lower/upper letters is not important (i.e. Uijo), however, in this instance the enumeration values are taken from an external source and thus it might be better to restrict them to the original (lower) case.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11855,7 +11855,7 @@ save_
 save_space_group_wyckoff.letter
 
     _definition.id                '_space_group_Wyckoff.letter'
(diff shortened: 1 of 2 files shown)


jamesrhester:
Nice work, thank you.

Issue #252, opened 2021-09-17
cif_core: capitalisation of certain data names


vaitkus:
It is useful to start off this discussion by stressing that data name capitalisation is purely a cosmetic issue and data names in any form can be legitimately used regardless of the form they appear in the dictionary. That being said, the `CIF_CORE` dictionary contains several data names that seemingly do not follow the capitalisation recommendations specified in rule 2.1.11 of the DDLm dictionary style guide [1]. The data names in question contain initialisms ("DOI", "ADP", "RMS", "IUCr", "CCDC"), acronyms ("ORCID", "InChI", "IUPAC") as well as chemical element names ("Cu", "Mo") that would normally be capitalised in the English language. Historically, these data names have not been capitalised and it is unclear to me if this should be changed.

This question was inspired by a related question from the `TopoCIF` dictionary development -- should acronyms like SMILES, InChI and IUPAC be capitalised given that they are not capitalised in the `CIF_CORE` dictionary?

[1] https://github.com/COMCIFS/comcifs.github.io/blob/master/accepted/ddlm_dictionary_style_guide.md

The (hopefully) full list of affected data names:
```
_atom_site.adp_type
_atom_site.refinement_flags_adp
_atom_type_scat.dispersion_imag_cu
_atom_type_scat.dispersion_imag_mo
_atom_type_scat.dispersion_real_cu
_atom_type_scat.dispersion_real_mo
_audit_author.id_orcid
_audit.block_doi
_audit_support.funding_organization_doi
_chemical_formula.iupac
_chemical.identifier_inchi
_chemical.identifier_inchi_key
_chemical.identifier_inchi_version
_citation.doi
_database_code.depnum_ccdc_archive
_database_code.depnum_ccdc_fiz
_database_code.depnum_ccdc_journal
_database.dataset_doi
_journal.paper_doi
_publ_author.id_iucr
_publ_author.id_orcid
_publ_contact_author.id_iucr
_publ_contact_author.id_orcid
_refine_diff.density_rms
_refine_diff.density_rms_su
```

jamesrhester:
I have no objection to regularising capitalisation for these data names. It has no practical effect so is low-impact.

PR #257, opened 2021-09-24
Update data name capitalisation

vaitkus:
This PR resolves issue #252.


jamesrhester:
Thanks for the speedy fix.

PR #267, opened 2021-10-19
Issue template

### LINKED ISSUE #259: Add a template for suggesting a new data name
Github has issue templates, we should prepare one for suggesting a new data name.

jamesrhester:
A first attempt at an issue template for suggesting new definitions.

Closes #259.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-definition.yml
+++ b/.github/ISSUE_TEMPLATE/new-definition.yml
@@ -12,7 +12,7 @@ body:
           attributes:
                   label: definition
                   description: "Please define the meaning of the proposed data name. You are welcome to use references as necessary"
-                  placeholder: Example: the number of symmetrically-equivalent orientations of the group
+                  placeholder: "Example: the number of symmetrically-equivalent orientations of the group"
           validations:
                   required: true
         - type: input
--- a/.github/ISSUE_TEMPLATE/new-definition.yml
+++ b/.github/ISSUE_TEMPLATE/new-definition.yml
@@ -2,8 +2,7 @@ name: Propose New Definition
 description: Propose a new definition for the Core dictionary
 title: "[Definition]: "
 labels: ["enhancement"]
-assignees:
-        - jamesrhester
+assignees: jamesrhester
 body:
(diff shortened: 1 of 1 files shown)


jamesrhester:
The aim with this issue template is to strike a balance between DDLm rigour and the likelihood that the submitter may not be that familiar with CIF terminology. So I tried to be careful not to overload the submitter with technical questions and terminology, while still obtaining as much useful information as possible.

> Some additional suggestions:
> 
>  1. It would be very useful to have fields for example values and example explanations. These could potentially even be made mandatory.

I agree that this is a good idea, I'll update the pull request with these.

>  2. Personally,  I would prefer to have "Data name" as the first input field and maybe even make it a mandatory one. Of course, one might not immediately known the exact final data name, but even having a placeholder value would be useful during the discussion.

I took the approach that, for the submitter, they are much more likely to have an idea of the definition of the item they want than be able to choose even a temporary data name. By putting the definition first, it is more welcoming, whereas a data name could create a mental issue that distracts from what they do want to convey. Sort of like it can be surprisingly hard to choose a filename when saving a file!

>  3. The previous IUCr web form provided a field to suggest the category that the item should belong to. I assume that in this form a somewhat similar role is served by the "Looping" section (at least for looped categories). It might be worthwhile to add a Category field in this form as well, although, the current approach might already be sufficient or even more intuitive.

As explained above, I deliberately avoided CIF terminology like 'Category', whereas which loop somebody expects to see a data name in gives the same information and more intuitively. It doesn't cover Set categories of course but that can be even harder for a non-expert to work out. 

>  4. The suggested `Date` data type might be better replaced by `DateTime` since the DDLm reference dictionary 
> recommends using `DateTime` instead of `Date` for all new dictionaries. I guess this sentiment extends to new data items > as well.

Agreed.

>  5. The term "Compound" has multiple meanings and is thus not very intuitive. Maybe the section could be renamed to "Data structure" and the `Single` value removed? This way the default value ("None") could be interpreted as the DDLm default ("Single") and the interpretation of the remaining values would remain the same.

Sounds good, I'll do that and update the PR.

PR #274, opened 2021-10-25
Changed atomic labels to type 'Word' for conformance with DDL1

### LINKED ISSUE #245: Items have become case-insensitive in DDLm core relative to DDL1

jamesrhester:
Closes #245

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -18790,7 +18790,7 @@ save_publ_manuscript_incl_extra.item
     _type.purpose                 Encode
     _type.source                  Assigned
     _type.container               Single
-    _type.contents                Text
+    _type.contents                Tag
 
 save_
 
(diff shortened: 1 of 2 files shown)


jamesrhester:
So I've gone through the list and made a best guess as to the case-sensitivity of external identifiers. For the others I've tended to err on the side of case-sensitivity for conformance with DDL1.

vaitkus:
Great, thank you for the work.

One last questions -- wouldn't it be better to leave the content type of `_publ_manuscript_incl_extra.item` data item as `Tag` instead of changing it to `Word`? CIF tags (data names) are case-insensitive and the `Tag` content type provides some additional value validation (requires the item to start with an underscore).

jamesrhester:
My reading of the definition suggested that more than one tag could be included in this list. Actually 'Word' is also incorrect in that case, it should be just 'Text', I'll fix that now.

vaitkus:
Great, thank you for the multiple iterations of changes.

PR #276, opened 2021-11-04
Updated checking workflow to allow caching.

jamesrhester:
This pull request updates the 'layout' part of the checking workflow to cache the julia setup, saving over a minute due to avoiding downloading and precompilation. This also dependence on the previously-used action from the workflow.


jamesrhester:
Also while I was at it I changed the workflow trigger to avoid running when the .github directory is the only one affected.

PR #279, opened 2021-11-09
Added detailed explanation of 'Replace' if_dupl mode for importation.

### LINKED ISSUE #264: Handling loops during 'if_dupl' imports
A thorough reviewer of the DDLm chapter has noted that the behaviour of imports in the case of `if_dupl` = `Replace` needs clarification in the particular case where a single packet loop is presented as unlooped data items. If the importing definition has two data names from the category, and the imported definition has only one of those two, then it is possible that only the duplicated data name will be replaced, leading to incorrect results (as the unduplicated value presumably was relevant only to the previous, replaced value).

I propose explicitly stating that all data items in a category are replaced, or removed if not present in the imported definition, which I think is the intended behaviour. Any comments?

jamesrhester:
Closes #264

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -1204,7 +1204,13 @@ save_import_details.if_dupl
       _enumeration_set.detail
          Ignore
 ;
-         Ignore imported definitions if id conflict.
+         Ignore imported definitions if block identifiers match in "Full" mode.
+         Ignore imported attributes that match attributes already in the
+         importing definition in "Contents" mode.
+         
+         When importing in "Contents" mode, if the ignored attribute belongs to
+         a Loop category, all attributes from that category must be ignored to
+         avoid loop mismatches.
 ;
          Replace
 ;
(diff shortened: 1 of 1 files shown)


vaitkus:
Great work. I have slightly expanded the description to explicitly state that the two definitions are merged and not completely replaced by one another. The updated definition is slightly more verbose, but hopefully clearer.

However, I have noticed that the 'Ignore' state is also slightly ambiguous in regards to the 'Contents' import type. It can be interpreted as:
1) Skip the import of all attributes in case a single duplicate attribute is detected.
2) Skip the import of that specific duplicate attribute.

 Note, that if the 2) option is to be assumed, it may also lead to a similar loop inconsistency unless explicitly specified that a failed import of one attribute from a Loop category halts the import of other items from that category. In this case, it might also then be worthwhile to mention this is the IT Vol. G. chapter as well.

The definition of the entire save frame could also be slightly improved to reflect that it covers not only duplicate definitions, but also duplicate attributes in definitions in case of the "Contents" type import.

jamesrhester:
Good catch. I have updated accordingly.

PR #285, opened 2021-11-14
Update descriptions of items with implied constraints

vaitkus:
This PR updates the descriptions of the `_enumeration.default` and `_description_example.case` data items to explicitly list
all the constraints that they inherit from the defining item.


jamesrhester:
I have merged this. Do we consider the value '.' to be implicitly acceptable as a default value regardless of constraints?

vaitkus:
I would say yes. To my understanding, CIF special values `?`, `.` implicitly satisfy any constraints of an individual value.

PR #288, opened 2021-11-18
Fix for incorrect ordering of 2 su data names in dictionary.

jamesrhester:
This fixes the incorrect order in two su data names brought about by an incorrect pretty printer/linter. It also fixes the workflow which was necessary because of an update to the linter used to check layout.

Note that the automated check will fail until the workflow is fixed by this PR.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,7 +62,7 @@ jobs:
       - name: Install Julia packages
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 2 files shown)


vaitkus:
I have merged the PR, but the automated checks still fail. May be related to issues outlined in [1] and [2].

[1] https://github.com/jamesrhester/cif_dic_layout_check_action/pull/2
[2] https://github.com/jamesrhester/julia_cif_tools/issues/6#issuecomment-973181012

jamesrhester:
This failure was due to a cached version of the Julia environment being used. I will manually force a new cache to be generated - not sure if there is an automated way to make this happen.

vaitkus:
Thanks!

Issue #275, opened 2021-10-27
cif_core.dic: homogenise the usage of DOI data items


vaitkus:
The `CIF_CORE` dictionary currently contains 5 DOI data items with the following properties:

| Data name | Content type | Expression |
| :---              | :---                 | :---              |
| _audit.block_DOI  | Text | DOI only |
| _audit_support.funding_organization_DOI | Text | DOI with resolver |
| _citation.DOI | Text | DOI only |
| _database.dataset_DOI | Text | DOI only |
| _journal.paper_DOI | Code | DOI only (?) |

### Content type

The `Text` content type seems the more appropriate one for two reasons:
1. DOI names are case-insensitive, using ASCII case folding for comparison of text. That is, two DOIs that only differ in the case of non-ASCII characters may be two different identifiers [1].
2. DOI names may incorporate any printable characters from the Universal Character Set (UCS-2), including the space [2].

### Expression
Data items `_audit.block_DOI`, `_citation.DOI`, `_database.dataset_DOI` have explicit examples which show that the values are stand-alone DOI strings (e.g. 10.2210/pdb4hhb/pdb) while `_journal.paper_DOI` does not provide an explicit example, but has historically been used in the same way.

The `_audit_support.funding_organization_DOI` data item, on the other hand, explicitly requires a full URI and seems to have a good reason to do so. However, this seems a bit suboptimal since resolver addresses may change (e.g. dx.doi.org was used prior to doi.org) and it is not currently clear if any special symbols should be escaped in the DOI URL string as specified in the DOI handbook [3] or should this be done by the user of the data. I am not sure if such special symbols occur at all in the given DOI range (funding organisations), but generally double encoding or not encoding the values may lead to incorrect interpretation of the data.

### Suggestions

Based on the previously given information I propose the following:
1. Change the content type of `_journal.paper_DOI` from `Code` to `Text`.
2. Add an explicit example to `_journal.paper_DOI` that showcases that the values should be stand-alone DOI strings.
3. Do one of the following for the `_audit_support.funding_organization_DOI` data item:
  a) Explicitly specify that no URL encoding is needed.
  b) Explicitly specify that special characters must be encoded and change the data type to `Word` thus preventing the use of the space symbol (which should be encoded as "%20" anyway).
  c) Change the value from full URL to stand-alone DOI.  

I assume that 1) and 2) will raise no objections, but which (if any) approach of 3) would be most suitable?  

### References
[...]

PR #289, opened 2021-12-01
Homogenise DOI data items

vaitkus:
This PR changes the content type of the `_journal.paper_doi` data item from 'Code' to 'Text' and adds an example case as discussed in issue #275.

Note, that I used a fake resolvable DOI [1] in the example (10.5555/12345678) to avoid unwanted/unexpected traffic to any real publication that may arise from it being included in a publication. Please change it to a more relevant one if needed.

[1] https://www.crossref.org/blog/doi-like-strings-and-fake-dois/

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.1.0
-    _dictionary.date              2021-11-15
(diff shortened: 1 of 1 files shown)


vaitkus:
Also, please note that the `layout` check currently fails for all runs due to a slight misconfiguration, but all of the checks passed successfully on my machine.

jamesrhester:
The layout checking failure is just waiting for the latest version of CrystalInfoFramework.jl to update in the Julia general registry.

PR #296, opened 2022-04-01
More units

jamesrhester:
Add units used in imgCIF dictionary to main units list.

Update workflow to use later version of linting software.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,7 +91,7 @@ jobs:
                for file in main/*.dic
                do      
                   echo "Checking $file"
-                  julia -O0 ./julia_cif_tools/linter.jl $file cif_core/ddl.dic
+                  julia -O0 ./julia_cif_tools/linter.jl -i $pwd/main $file cif_core/ddl.dic
                   if [ $? != 0 ] 
                   then 
                     exit 1 ;
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -91,7 +91,7 @@ jobs:
                for file in main/*.dic
                do      
                   echo "Checking $file"
-                  julia -O0 ./julia_cif_tools/linter.jl -i $pwd/main $file cif_core/ddl.dic
+                  julia -O0 ./julia_cif_tools/linter.jl -i $PWD/cif_core $file cif_core/ddl.dic
                   if [ $? != 0 ] 
                   then 
                     exit 1 ;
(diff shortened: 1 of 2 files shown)


vaitkus:
Looks good, but I do have several question:

1. Could you please provide a link to the imgCIF dictionary file that these new values were taken from? There are several versions available [1, 2] and none of them seem to contain the newly added "photons per second" units.
2. Are these new additions intended to cover all the extra units that appear in imgCIF or only the select few?
3. Could the newly added "photons per second" unit be renamed to "photons_per_second" to have the same syntax as the rest of the already defined units (note the underscores).  

[1] https://github.com/COMCIFS/imgCIF/blob/master/cif_img.dic
[2] https://www.iucr.org/__data/iucr/cif/dictionaries/cif_img.dic

jamesrhester:
1. The "official" version of the imgCIF dictionary is inside the cbflib repository documentation directory [1]. The latest version would be 1.8.4 and note some important outstanding pull requests. The `photons per second` units are found in `_diffrn_radiation.beam_flux`.
2. These additions are intended to cover all units that appear in imgCIF
3. As we are trying to maintain interconvertibility with DDL2 imgCIF, it is probably best that we don't touch the contents of text items.

In general DDL2 imgCIF is maintained in a fairly git-unfriendly way (explicit version numbering in name, is provided together with software). I hope that when we have a DDLm imgCIF up and running we will be able to provide a more git-friendly experience while still allowing the DDLm version to be back-converted to DDL2. I am close to having an auto-generated DDLm version available.
  
[1] https://github.com/yayahjb/cbflib/tree/main/doc

jamesrhester:
I suggest we merge what we have as a "perfect" copy of the current DDL2 imgCIF dictionary, and any issues that arise we should raise within the DDL2 imgCIF repository. If and when those issues are resolved we can update our copy accordingly. We can also consider maintaining our own branch of the DDLm version that includes what we think should be fixed, but that runs the danger of creating conflicting definitions for the same data name.

vaitkus:
Acknowledged. I have added a description of the changes and merged in the PR.

I merged the PR even though some formatting errors in the `cif_core.dic` that were reported by the linter since these reports seem to be false-positive (enumeration values without any delimiters are reported as having mismatching delimiters).

vaitkus:
Thank you for the clarifications. I raised an issue which deals with the handling of CIF special values in regards to the style rules (https://github.com/COMCIFS/comcifs.github.io/issues/13) so it could be further properly discussed if needed.

PR #298, opened 2022-05-09
Defined basic data names to allow multi-block data sets

jamesrhester:
Please see the document multi-block-principles.md in the comcifs.github.io directory for information on
use of these new data names. Note that both new data names are already defined in mmCIF with identical roles.


jamesrhester:
Note failure of CI is due to flagging of key data names for `Set` categories and some style rule violations due to a bug in the linter, so I believe it is OK to go ahead and merge these updates.

vaitkus:
Just to double check -- maybe the `_exptl_crystal.id` data item should then also be exclusively included as the key of the  `EXPTL_CRYSTAL` category?

jamesrhester:
Yes, actually, that should be a key of that category.

Issue #294, opened 2021-12-16
Does the `CELL` category really belong in `EXPTL`?


jamesrhester:
The `Set` category `EXPTL` is described as: 

> data items used to specify the experimental work
>     prior to diffraction measurements.

The `Set` `CELL` category is formally a child of `EXPTL`, suggesting that cells are derived before the experiment is run. While this was true in the early 90s when using point detectors, essentially all experiments now derive the cell from the diffraction images and so `CELL` belongs more clearly in the `DIFFRN` category. However, `CELL` has data names allowing the cell measurement temperature, pressure, and wavelength to be specified separately to those specified in `DIFFRN`.

Where multiple temperatures are used in an experiment, multiple cells will be determined as they will differ by temperature. Seen in this light, there is an implicit child data name of `_diffrn.id` in `CELL` which is only allowed to be skipped due to the `Set` nature of the `CELL` category. This further implies that `CELL` should be a child of `DIFFRN`, although it is true that the parent-child relationship of `Set` categories has no practical semantic implications.

So, assuming that we need to link `CELL` with `DIFFRN` in both a machine-readable and human-sensible manner, here are some steps to take:

1. Change the parent of `CELL` to `DIFFRN`

This has no noticeable effect on practical use but would improve consistency for very attentive human readers.

2. Add a child data name of `_diffrn.id` to `CELL` and child categories, including `CELL_MEASUREMENT`

This is a machine-readable way of stating that a particular cell relates to a particular set of diffraction conditions, and would be a sufficient basis for validation software to raise flags if values are different, without outright disallowing it.

3. Alias `_cell_measurement.{temperature/pressure/radiation}` to equivalent data names in `DIFFRN`

In combination with (2), this would make contradictory values semantically invalid. We would need to check a large corpus of all CIF files e.g. COD for conformance to see if this assumption holds.

4. Deprecate `_cell_measurement.{temperature/pressure/radiation}` in favour of the `DIFFRN` equivalents, or visa versa

This seems unlikely to make any difference as these data names are widely encoded into current reading and writing software.

My preference would be to do 1, 2 and 3, providing a COD check does not find any disagreements in existing files.

vaitkus:
[...]

PR #299, opened 2022-05-25
Moved CELL to DIFFRN category, deprecated some items.

jamesrhester:
This addresses items discussed in issue #294.
 
The CELL category parent is now DIFFRN.
Pressure, temperature items in CELL_MEASUREMENT deprecated
in favour of identical items in DIFFRN.

Pointers to `_diffrn.id` added to CELL etc. to allow multiple
cells to be presented where multiple diffraction
conditions are used.

Note that in the now unusual case that cell measurement conditions are different to diffraction conditions, a separate data block with different `_diffrn.id` can be created to describe those measurement conditions, and a pointer to that `_diffrn.id` used. Therefore the `_cell_measurement.pressure,temperature` items are now completely unnecessary.

The diff looks quite large because the CELL category and all children have been moved around.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.2.0
-    _dictionary.date              2022-05-09
+    _dictionary.date              2022-05-25
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.1.0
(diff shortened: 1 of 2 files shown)


vaitkus:
> This addresses items discussed in issue https://github.com/COMCIFS/cif_core/issues/294.
>
> The CELL category parent is now DIFFRN.
> Pressure, temperature items in CELL_MEASUREMENT deprecated
> in favour of identical items in DIFFRN.

Great, thank you! However, must the `_cell_measurement.{pressure,temperature}` items be formally deprecated? They still serve a distinct purpose in some older CIFs where the `CELL_MEASUREMENT` and `DIFFRN` values do differ (see https://github.com/COMCIFS/cif_core/issues/294#issuecomment-999185470). I much more like the approach suggested in the final comment of issue #294 (https://github.com/COMCIFS/cif_core/issues/294#issuecomment-1019637363):

1. Shift `CELL/CELL_MEASUREMENT(_REFLN)` categories to have `DIFFRN` as the parent. _(This is fully implemented in the current PR.)_
2. Clarify the `CELL_MEASUREMENT` definitions to emphasise that the temp/pressure items relate only to measurements performed separately to the diffraction experiment. _(This is partially implemented in the current PR.)_
3. Add dREL methods to the  `_cell_measurement.{temp,pressure}` that take their default values from the equivalent `DIFFRN` data items. _(This is not implemented in the current PR.)_

Point 2 could be improved by removing the deprecation notice and the `_definition_replaced.by` attribute and instead adding the suggested disclaimer. For example, the updated description of `_cell_measurement.pressure` could read:

```
    This data item should only be used if the unit-cell parameters were
    measured separately from the diffraction experiment. Otherwise,
    the `_diffrn.ambient_pressure` data item should be used instead.

    The pressure at which the unit-cell parameters were measured
    (not the pressure used to synthesize the sample).
```

Point 3 could be implemented by adding the corresponding methods. Also, care should be taken to only assign `*_su` data items values if the associated measurand data item is not explicitly provided. For example, if a file contains `_diffrn.ambient_pressure`, `_diffrn.ambient_pressure_su` and `_cell_measurement.pressure` data items, the missing  `_cell_measurement.pressure_su` value MUST NOT be automatically assigned from  `_diffrn.ambient_pressure_su` since this may lead to incorrect data (though it should still be ok to take this value if both  `_cell_measurement.pressure` and  `_cell_measurement.pressure_su` are missing).

What do you think?

> Note that in the now unusual case that cell measurement conditions are different to diffraction conditions, a separate data block with different _diffrn.id can be created to describe those measurement conditions, and a pointer to that _diffrn.id used. Therefore the _cell_measurement.pressure,temperature items are now completely unnecessary.

Could you please provide a short example of how this would look? I get the general idea of this, but do not yet fully understand how the `_cell_measurement.{pressure,temperature}` data items could be completely avoided.

> The diff looks quite large because the CELL category and all children have been moved around.

In this particular case the diff provided by the GitHub interface is indeed not very convenient to read, however, this can be easily circumvented by downloading the file and using a different diff client or simply reordering the file offline.

Also, the `ddlm` step of the `CIFSyntaxandStyleCheck` action is currently failing due to the missing `_definition_replaced.id` attributes. The new version of the validator will have the option to allow implicit key values and will thus suppress these messages. Alternatively, if the changes proposed in point 2 are implemented, the messages should go away on their own with the removal of the newly added `_definition_replaced.id` attributes. Finally, if the want to quickly silence these messages we can either add the missing `_definition_replaced.id` attributes or temporarily patch the action to filter out these messages.

rowlesmr:
I second the call for an example.

.

Also, are others able to modify this PR? I'm getting authentication issues in github desktop, and I don't know if they're me, the program, or the repo.

jamesrhester:
OK, will work up an example.

jamesrhester:
OK, so below is a vanilla example of a CIF file with two data blocks, which is merged according to the presence of a `_diffrn.id` child key data name in the category as defined by the dictionary. Not sure if this is the example you were after. Notes:

1.  Whenever the child data name of `_diffrn.id` is present in the dictionary, it has been explicitly added to the merged loops.
2. If information in a category is precisely duplicated in both files, then it is simply repeated in the merged file. Otherwise it is an error.
3. This is a demonstration only of how the merging works. More categories will need some thought, in particular the symmetry could change with temperature and/or pressure. I propose addressing this in a separate PR, as this PR considers only the relationship of `CELL` to `DIFFRN`.

```
#\#CIF_2.0
#
# A test file with two data blocks that should be mergeable
#
# Represents an experiment conducted on two crystals, one
# at one temperature and pressure and Cu K alpha, the
# other at a different temperature and pressure and Mo Ka
#
data_saly2_all_aniso

_chemical_formula.sum                 'C7 H6 O3'
_chemical_formula.weight               132.07

_symmetry.cell_setting                 monoclinic
_symmetry.space_group_name_H-M         P_1_21/a_1
_symmetry.space_group_name_Hall        -p_2yab

loop_
_space_group_symop.operation_xyz
  '+x,+y,+z'  '1/2-x,1/2+y,-z'  '-x,-y,-z'  '1/2+x,1/2-y,+z'
 
_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_exptl_crystal.id                      second
_exptl_crystal.density_diffrn          1.444
_exptl_crystal.density_meas            1.234

_diffrn.ambient_temperature            293
_diffrn.ambient_pressure               10
_diffrn_radiation.type                'Cu K\a'
_diffrn_radiation_wavelength.value     1.5418
_diffrn_radiation.probe                x-ray
_reflns.apply_dispersion_to_Fcalc       Yes

loop_
_atom_type.symbol
_atom_type.atomic_mass
_atom_type.number_in_cell
  O       15.999   12 
  C       12.011   28
  H       1.008    24


loop_
_atom_type_scat.symbol
_atom_type_scat.dispersion_real
_atom_type_scat.dispersion_imag
_atom_type_scat.source
  O     .047  .032  'Int Tables Vol IV Tables 2.2B and 2.3.1'
  C     .017  .009  'Int Tables Vol IV Tables 2.2B and 2.3.1'
  H      0  0  'Int Tables Vol IV Tables 2.2B and 2.3.1'

 
loop_
_atom_site.label
_atom_site.fract_x
_atom_site.fract_y
_atom_site.fract_z
_atom_site.U_iso_or_equiv
_atom_site.adp_type
_atom_site.occupancy

  o1  .5505(5)  .6374(5)  .1605(11)  .035(3)  Uani   1.00000 
  o2  .4009(5)  .5162(5)  .2290(11)  .033(3)  Uani   1.00000 
  o3  .2501(5)  .5707(5)  .6014(13)  .043(4)  Uani   1.00000
  c1  .4170(7)  .6930(8)  .4954(15)  .029(4)  Uani   1.00000
  c2  .3145(7)  .6704(8)  .6425(16)  .031(5)  Uani   1.00000 
  c3  .2789(8)  .7488(8)  .8378(17)  .040(5)  Uani   1.00000 
  c4  .3417(9)  .8529(8)  .8859(18)  .045(6)  Uani   1.00000 
  c5  .4445(9)  .8778(9)  .7425(18)  .045(6)  Uani   1.00000
  c6  .4797(8)  .7975(8)  .5487(17)  .038(5)  Uani   1.00000 
  c7  .4549(7)  .6092(7)  .2873(16)  .029(4)  Uani   1.00000 

loop_                            
 _refln.index_h                   
 _refln.index_k                   
 _refln.index_l                   
 _refln.F_meas                    
 _refln.F_calc                    
    0   0   3  26.557  23.993
    0   1   1  33.109  32.058
    1   1   1   9.008   6.604

data_next_block
 
_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_exptl_crystal.id                      first
_exptl_crystal.density_diffrn          1.444
_exptl_crystal.density_meas            1.234

_diffrn.ambient_temperature            293
_diffrn.ambient_pressure               10

_diffrn_radiation.type                'Mo K\a'
_diffrn_radiation_wavelength.value     0.701
_diffrn_radiation.probe                x-ray

_reflns.apply_dispersion_to_Fcalc       Yes

# These are all dummy values so that they look different
# from the previous block. Final digit is '6'

loop_
_atom_site.label
_atom_site.fract_x
_atom_site.fract_y
_atom_site.fract_z
_atom_site.U_iso_or_equiv
_atom_site.adp_type
_atom_site.occupancy

  o1  .5506(5)  .6376(5)  .1606(11)  .035(3)  Uani   1.00000 
  o2  .4006(5)  .5166(5)  .2296(11)  .033(3)  Uani   1.00000 
  o3  .2506(5)  .5706(5)  .6016(13)  .043(4)  Uani   1.00000
  c1  .4176(7)  .6936(8)  .4956(15)  .029(4)  Uani   1.00000
  c2  .3146(7)  .6706(8)  .6426(16)  .031(5)  Uani   1.00000 
  c3  .2786(8)  .7486(8)  .8376(17)  .040(5)  Uani   1.00000 
  c4  .3416(9)  .8526(8)  .8856(18)  .045(6)  Uani   1.00000 
  c5  .4446(9)  .8776(9)  .7426(18)  .045(6)  Uani   1.00000
  c6  .4796(8)  .7976(8)  .5486(17)  .038(5)  Uani   1.00000 
  c7  .4546(7)  .6096(7)  .2876(16)  .029(4)  Uani   1.00000 

loop_                            
 _refln.index_h                   
 _refln.index_k                   
 _refln.index_l                   
 _refln.F_meas                    
 _refln.F_calc                    
    0   0   3  26.556  23.996
    0   1   1  33.106  32.056
    1   1   1   9.006   6.606
```

Merged:

```
_symmetry.space_group_name_hall	-p_2yab
_symmetry.space_group_name_h-m	P_1_21/a_1
    loop_
      _space_group_symop.operation_xyz
         +x,+y,+z
         1/2-x,1/2+y,-z
         -x,-y,-z
         1/2+x,1/2-y,+z
 loop_
      _cell.angle_alpha
      _cell.angle_beta
      _cell.angle_gamma
      _cell.diffrn_id
      _cell.formula_units_z
      _cell.length_a
      _cell.length_b
      _cell.length_c
      _cell.volume
         90.00000  90.8331(5)  90.00000  1  4  11.520(12)  11.210(11)  4.920(5)
         635.3(11)
         90.00000  90.8331(5)  90.00000  2  4  11.520(12)  11.210(11)  4.920(5)
         635.3(11)
    loop_
      _diffrn.ambient_pressure
      _diffrn.ambient_temperature
      _diffrn.id
         10                                293                                1
         10                                293                                2
    loop_
      _diffrn_radiation.diffrn_id
      _diffrn_radiation.probe
      _diffrn_radiation.type
         1                            x-ray                            'Cu K\a'
         2                            x-ray                            'Mo K\a'
    loop_
      _exptl_crystal.density_diffrn
      _exptl_crystal.density_meas
      _exptl_crystal.id
         1.444                           1.234                           second
         1.444                           1.234                           first
    loop_
      _diffrn_radiation_wavelength.diffrn_id
      _diffrn_radiation_wavelength.id
      _diffrn_radiation_wavelength.value
         1                               1                               1.5418
         2                               2                               0.701
    loop_
      _atom_type.atomic_mass
      _atom_type.number_in_cell
      _atom_type.symbol
         15.999                              12                              O
         12.011                              28                              C
         1.008                               24                              H
    loop_
      _atom_type_scat.dispersion_imag
      _atom_type_scat.dispersion_real
      _atom_type_scat.source
      _atom_type_scat.symbol
         .032      .047      'Int Tables Vol IV Tables 2.2B and 2.3.1'      O
         .009      .017      'Int Tables Vol IV Tables 2.2B and 2.3.1'      C
         0         0         'Int Tables Vol IV Tables 2.2B and 2.3.1'      H

    loop_
      _chemical_formula.sum
      _chemical_formula.weight
         'C7 H6 O3'               132.07
    loop_
      _reflns.apply_dispersion_to_fcalc
         Yes
    loop_
      _symmetry.cell_setting
         monoclinic
    loop_
      _atom_site.adp_type
      _atom_site.diffrn_id
      _atom_site.fract_x
      _atom_site.fract_y
      _atom_site.fract_z
      _atom_site.label
      _atom_site.occupancy
      _atom_site.u_iso_or_equiv
         Uani   1   .5505(5)   .6374(5)   .1605(11)   o1   1.00000   .035(3)
         Uani   1   .4009(5)   .5162(5)   .2290(11)   o2   1.00000   .033(3)
         Uani   1   .2501(5)   .5707(5)   .6014(13)   o3   1.00000   .043(4)
         Uani   1   .4170(7)   .6930(8)   .4954(15)   c1   1.00000   .029(4)
         Uani   1   .3145(7)   .6704(8)   .6425(16)   c2   1.00000   .031(5)
         Uani   1   .2789(8)   .7488(8)   .8378(17)   c3   1.00000   .040(5)
         Uani   1   .3417(9)   .8529(8)   .8859(18)   c4   1.00000   .045(6)
         Uani   1   .4445(9)   .8778(9)   .7425(18)   c5   1.00000   .045(6)
         Uani   1   .4797(8)   .7975(8)   .5487(17)   c6   1.00000   .038(5)
         Uani   1   .4549(7)   .6092(7)   .2873(16)   c7   1.00000   .029(4)
         Uani   2   .5506(5)   .6376(5)   .1606(11)   o1   1.00000   .035(3)
         Uani   2   .4006(5)   .5166(5)   .2296(11)   o2   1.00000   .033(3)
         Uani   2   .2506(5)   .5706(5)   .6016(13)   o3   1.00000   .043(4)
         Uani   2   .4176(7)   .6936(8)   .4956(15)   c1   1.00000   .029(4)
         Uani   2   .3146(7)   .6706(8)   .6426(16)   c2   1.00000   .031(5)
         Uani   2   .2786(8)   .7486(8)   .8376(17)   c3   1.00000   .040(5)
         Uani   2   .3416(9)   .8526(8)   .8856(18)   c4   1.00000   .045(6)
         Uani   2   .4446(9)   .8776(9)   .7426(18)   c5   1.00000   .045(6)
         Uani   2   .4796(8)   .7976(8)   .5486(17)   c6   1.00000   .038(5)
         Uani   2   .4546(7)   .6096(7)   .2876(16)   c7   1.00000   .029(4)
  
    loop_
      _refln.diffrn_id
      _refln.f_calc
      _refln.f_meas
      _refln.index_h
      _refln.index_k
      _refln.index_l
         1          23.993          26.557          0          0          3
         1          32.058          33.109          0          1          1
         1          6.604           9.008           1          1          1
         2          23.996          26.556          0          0          3
         2          32.056          33.106          0          1          1
         2          6.606           9.006           1          1          1
   
```

jamesrhester:
Here is a further example showing how I propose the cell measurement conditions would now be presented, if they occur in different conditions to the main experiment. I suggest this is better, as there is no need to duplicate the environmental definitions that are already present in `DIFFRN`. This is why I prefer deprecation. I note that if the data set contains more than one `_diffrn.id`, it will be necessary to explicitly provide `_cell_measurement.diffrn_id` unless we make it a key data name - which I have done in the PR.

```
#\#CIF_2.0
#
# A test file with two data blocks
#
# Represents an experiment where the cell measurement
# was performed under different conditions to the data
# collection.
#
data_main_collection
 
_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_cell_measurement.diffrn_id     2 # look elsewhere
_cell_measurement.theta_max  50.5
_cell_measurement.theta_min 10.1
_cell_measurement.reflns_used 20 

_diffrn.id                                  1
_diffrn.ambient_temperature            293
_diffrn.ambient_pressure               10
_diffrn_radiation.type                'Cu K\a'
_diffrn_radiation_wavelength.value     1.5418
_diffrn_radiation.probe                x-ray

# other results follow, atom sites etc.

data_cell_measurement
 
_diffrn.id                                    2
_diffrn.ambient_temperature            290  #aircon switched on
_diffrn.ambient_pressure               10

_diffrn_radiation.type                'Mo K\a' #moly tube inserted
_diffrn_radiation_wavelength.value     0.701
_diffrn_radiation.probe                x-ray

**Comment by rowlesmr** [comment]
I haven't read the dictionary changes, but the way the `data_main_collection` and `data_cell_measurement` examples are shown, I think clearly show what is going on. There are two separate experiments being conducted, which is now shown explicitly.

**Comment by vaitkus** [comment]
Thanks, things are clearer now. The first merged example has a slight mistake as both experiments are assigned the same temperature, but I do get the main idea.

I do have a few notes on the second example. One thing that is being lost in this refactoring is the explicit declaration of the experiment's purpose. That is, it is no longer clear just by looking at the second data block that it describes a cell measurement (this is only evident from `_cell_measurement.diffrn_id` value in the first data block). This could probably be resolved by introducing an optional item with enumeration values, e.g. `_diffrn.measurand`: `lattice`, `coordinates`, `lattice_and_coordinates`, etc. This would also make things clearer when the data from the two blocks is looped together. Would this make sense?

Also, I see that some items related to the cell measurement (e.g. `_cell_measurement.reflns_used`) were placed in the first data block, although the cell measurement experiment is provided in the second block. I guess it is implied by the `_cell_measurement.diffrn_id` value that they are associated with a different experiment and can thus be freely moved from one block to the other (e.g. during the merging/splitting of the data blocks)?

**Comment by jamesrhester** [comment]
> I do have a few notes on the second example. One thing that is being lost in this refactoring is the explicit declaration of the experiment's purpose. That is, it is no longer clear just by looking at the second data block that it describes a cell measurement (this is only evident from `_cell_measurement.diffrn_id` value in the first data block). This could probably be resolved by introducing an optional item with enumeration values, e.g. `_diffrn.measurand`: `lattice`, `coordinates`, `lattice_and_coordinates`, etc. This would also make things clearer when the data from the two blocks is looped together. Would this make sense?

I don't think that this is going to work, as the referenced data block may (in theory) contain a complete experiment (e.g. at a different wavelength), the cell parameters of which were used for the other block. If all the block contains are diffraction conditions, then that speaks for itself: the block is a description of diffraction conditions, and to see where they were used you find any blocks that refer to that `_diffrn.id`.
> 
> Also, I see that some items related to the cell measurement (e.g. `_cell_measurement.reflns_used`) were placed in the first data block, although the cell measurement experiment is provided in the second block. I guess it is implied by the `_cell_measurement.diffrn_id` value that they are associated with a different experiment and can thus be freely moved from one block to the other (e.g. during the merging/splitting of the data blocks)?

This is an important point and exposes an error in my plan. As `_cell_measurement.diffrn_id` is a key data name, it can be elided if `_diffrn.id` is provided in the same data block. This therefore *cannot* have the meaning of "the diffraction conditions under which the cell was measured" but instead means "the diffraction conditions under which the diffraction experiment to which these cell measurements apply was measured". We should introduce a further data name "_cell_measurement.condition_id" which points to the `_diffrn.id` of the cell measurement conditions and defaults to `_cell_measurement.diffrn.id`. I will update the PR and example above.

**Comment by jamesrhester** [comment]
So the latest PR introduces `_cell_measurement.condition_id` and clarifies the meaning of the `_diffrn.id` data pointers to mean "the diffraction experiment to which this cell applies", rather than being "the conditions under which this cell was measured". 

Below is an updated example. Note that child data names of `_diffrn.id` are implicitly present in `CELL` and `CELL_MEASUREMENT` categories.
```
#\#CIF_2.0
#
# A test file with two data blocks
#
# Represents an experiment where the cell measurement
# was performed under different conditions to the data
# collection.
#
data_main_collection
 
_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_cell_measurement.condition_id     2 # look elsewhere
_cell_measurement.theta_max  50.5
_cell_measurement.theta_min 10.1
_cell_measurement.reflns_used 20 

_diffrn.id                                  1
_diffrn.ambient_temperature            293
_diffrn.ambient_pressure               10
_diffrn_radiation.type                'Cu K\a'
_diffrn_radiation_wavelength.value     1.5418
_diffrn_radiation.probe                x-ray

# other results follow, atom sites etc.

data_cell_measurement
 
_diffrn.id                                    2
_diffrn.ambient_temperature            290  #aircon switched on
_diffrn.ambient_pressure               10

_diffrn_radiation.type                'Mo K\a' #moly tube inserted
_diffrn_radiation_wavelength.value     0.701
_diffrn_radiation.probe                x-ray
```

**Comment by vaitkus** [comment]
I am not sure that I completely get the need for the `_cell_measurement.condition_id` data item. Is it really a problem if the `_cell_measurement.diffrn_id` data item value gets automatically elided from `_diffr.id`? Consider the following scenarios:
1. Cell and coordinates are measured during the same experiment. In this case, it doesn't matter if the `_cell_measurement.diffrn_id` is elided from `_diffr.id` or given explicitly since the two ids must match (it's the same experiment).
2. Cell and coordinates are measured during different experiments. In this case the only correct approach is to explicitly provide the id of the other data block since automatic value resolution fails in both of the suggested approaches. If an explicit  `_cell_measurement.diffrn_id` value is not given either `_cell_measurement.diffrn_id`  ends up equal to `_diffr.id` due to it being elided (approach 1) or  `_cell_measurement.condition_id`  ends up equal to `_diffr.id`  due to dREL (approach 2).

If we do decide to keep  `_cell_measurement.condition_id`, maybe it could be renamed to something less specific as to not tie it only specifically to measurement conditions, but rather to the experiment itself, for example `_cell_measurement.exptl_diffr_id`?

Also, I see that `_cell_measurment.condition_id` is explicitly linked to `_diffrn.id` using `_name.linked_item_id`. This will work properly when all of the information is provided in a single data block (merged/looped approach), but under a multi data block approach this will require the validation software to know that the `_diffrn.id` values should be searched for in other data blocks as well. Will the requirement for such behaviour be somehow explicitly marked in the CIF file itself (e.g. by using `_audit.formalist` (https://github.com/COMCIFS/comcifs.github.io/blob/master/draft/audit.formalism_proposal.md))? Furthermore, I somewhat recall similar multi-block relations being written up in the powder dictionary, but weren't the explicit links removed in those cases? Maybe @rowlesmr knows which issue/PR I am referring to?

**Comment by jamesrhester** [comment]
> I am not sure that I completely get the need for the _cell_measurement.condition_id data item. Is it really a problem if the _cell_measurement.diffrn_id data item value gets automatically elided from _diffr.id?

No, it is not a problem, it will normally be elided, and only exists to associate the cell measurement information to the experimental data collection it applies to. Note that `_cm.diffrn_id` and `_cm.condition_id` have different meanings, so we need both. `_cm.diffrn_id` refers to the data collection for which these cell parameters were *used*, e.g. to orient a crystal and point detector to measure a spot, or to calculate the position to extract spots from images. `_cm.condition_id` refers to the environmental conditions under which the cell itself was measured. The only time `_cm.condition_id` needs to be explicitly provided is if these conditions are different.

> Also, I see that _cell_measurement.condition_id is explicitly linked to _diffrn.id using _name.linked_item_id. This will work properly when all of the information is provided in a single data block (merged/looped approach), but under a multi data block approach this will require the validation software to know that the _diffrn.id values should be searched for in other data blocks as well. Will the requirement for such behaviour be somehow explicitly marked in the CIF file itself (e.g. by using _audit.formalism)

In general, comprehensive validation software will need to understand multi-block presentations (or refuse to do it!). Data blocks are invisible to DDLm, with the exception of the `Set` category. At this point in the development of DDLm, we have made/are making sure that data block merging into an overall single block is conceptually unambiguous and consistent, as per my example above.

**Comment by vaitkus** [comment]
> No, it is not a problem, it will normally be elided, and only exists to associate the cell measurement information to the experimental data collection it applies to. Note that `_cm.diffrn_id` and `_cm.condition_id` have different meanings, so we need both. `_cm.diffrn_id` refers to the data collection for which these cell parameters were _used_, e.g. to orient a crystal and point detector to measure a spot, or to calculate the position to extract spots from images. `_cm.condition_id` refers to the environmental conditions under which the cell itself was measured. The only time `_cm.condition_id` needs to be explicitly provided is if these conditions are different.

But will there ever be a situation under this approach where `_cm.diffrn_id` will be different from the `_diffrn.id` of the data block that contains the `_cm.diffrn_id`? Intuitively it seems correct to always assume that the cell parameters given in a data block were used to collect the rest of the data in the data block.

> In general, comprehensive validation software will need to understand multi-block presentations (or refuse to do it!). Data blocks are invisible to DDLm, with the exception of the `Set` category. At this point in the development of DDLm, we have made/are making sure that data block merging into an overall single block is conceptually unambiguous and consistent, as per my example above.

Understood. This quite differs from the way we currently generally do things in the COD (one data block - one structure) so properly recognizing files that conform to the DDLm will be of great importance to us. Hopefully, the `AUDIT_CONFORM` category will be more widely utilised by modern CIF-producing software.

**Comment by jamesrhester** [comment]
> But will there ever be a situation under this approach where _cm.diffrn_id will be different from the _diffrn.id of the data block that contains the _cm.diffrn_id? Intuitively it seems correct to always assume that the cell parameters given in a data block were used to collect the rest of the data in the data block.

The reason for going to the trouble of creating the possibility of these conditions being different is in response to 

> Great, thank you! However, must the _cell_measurement.{pressure,temperature} items be formally deprecated? They still serve a distinct purpose in some older CIFs where the CELL_MEASUREMENT and DIFFRN values do differ (see https://github.com/COMCIFS/cif_core/issues/294#issuecomment-999185470).

which I interpreted as asking for the capability to be preserved.

**Comment by vaitkus** [comment]
Sorry, I was a bit unclear with my comments. There is definitely the need to be able to specify that these conditions are different so thank you for accommodating this. My comment was aimed at understanding the semantics of the `_cm.conditions_id` and `_cm.diffrn_id` data items.

I'm not sure that I completely get it, but my current rationalisation for the introduction of `_cm.conditions_id` would be this:

* Data items from the `CM` category (e.g. `_cell_measurement.reflns_used`) may have different values in different data blocks (i.e. the main data collection block and the cell measurement block).
* Although the `CM` category is currently defined as a `Set`, under the new approach it can be looped if the two data blocks are merged. Due to this, the `_cell_measurement.diffrn_id` must uniquely identify the diffraction measurements rather than inform which diffraction measurement was used to get the cell data.

I tried to rework your usage example along the lines of these statements:
```
#\#CIF_2.0
#
# A test file with two data blocks
#
# Represents an experiment where the cell measurement
# was performed under different conditions to the data
# collection.
#
data_main_collection
 
_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_cell_measurement.condition_id         2 # look elsewhere
_cell_measurement.theta_max            50.5
_cell_measurement.theta_min            10.1
_cell_measurement.reflns_used          20 

_diffrn.id                                  1
_diffrn.ambient_temperature            293
_diffrn.ambient_pressure               10
_diffrn_radiation.type                'Cu K\a'
_diffrn_radiation_wavelength.value     1.5418
_diffrn_radiation.probe                x-ray

# other results follow, atom sites etc.

data_cell_measurement
 
_diffrn.id                             2
_diffrn.ambient_temperature            290  #aircon switched on
_diffrn.ambient_pressure               10

_diffrn_radiation.type                'Mo K\a' #moly tube inserted
_diffrn_radiation_wavelength.value     0.701
_diffrn_radiation.probe                x-ray

# Some items from the cell measurement category have different
# values from those provided in the main data collection block.
# _cell_measurement.condition_id       2 # look within
_cell_measurement.theta_max            50.4
_cell_measurement.theta_min            10.0
_cell_measurement.reflns_used          30

# A dataset merged from the first two.
data_merged

_cell.length_a                         11.520(12)
_cell.length_b                         11.210(11)
_cell.length_c                         4.920(5)
_cell.angle_alpha                      90.00000
_cell.angle_beta                       90.8331(5)
_cell.angle_gamma                      90.00000
_cell.volume                           635.3(11)
_cell.formula_units_Z                  4

_cell_measurement.diffrn_id
_cell_measurement.condition_id
_cell_measurement.theta_max
_cell_measurement.theta_min
_cell_measurement.reflns_used
1 2 50.5 10.1 20
2 2 50.4 10.0 30

_diffrn.id
_diffrn.ambient_temperature
_diffrn.ambient_pressure
1 293 10
2 290 10

loop_
_diffrn_radiation.diffrn_id
_diffrn_radiation.type
_diffrn_radiation_wavelength.value
_diffrn_radiation.probe
1 'Cu K\a' 1.5418 x-ray
2 'Mo K\a' 0.701 x-ray
```

The third data block (`merged`) represents the merging of the first two data blocks. Am I thinking in the right direction?

jamesrhester:
> I'm not sure that I completely get it, but my current rationalisation for the introduction of _cm.conditions_id would be this:

>    Data items from the CM category (e.g. _cell_measurement.reflns_used) may have different values in different data blocks (i.e. the main data collection block and the cell measurement block).
    Although the CM category is currently defined as a Set, under the new approach it can be looped if the two data blocks are merged. Due to this, the _cell_measurement.diffrn_id must uniquely identify the diffraction measurements rather than inform which diffraction measurement was used to get the cell data.

Yes. So the `CM` data names that appear in the second block in your example refer to the cell measurements that gave the
cell used for any data collection that appears in the second data block. The example could make sense if a group of reflections was measured under the conditions specified in the second data block, and only a subset of 20 of them were used to calculate the cell that was then used to guide the data collection in the first data block. The whole group of 30 were used to calculate the cell that was used to guide the data collection in the second data block. This might even have made some sense in the 1980s?

It would also be entirely reasonable to have no `CM` data names in the second data block in the example, so that the second data block is used purely to state the environmental conditions (only `diffrn` data names).

vaitkus:
Ok, thank you for the explanation. To move things forward, I suggest that we sync the branch with the current version of the main branch and then add the examples generated during this discussion as separate CIF files in the examples/ directory. As a bonus, for anyone else wanting to review this PR before the merge, the example files would provide a much clearer view of the whole approach.

jamesrhester:
I will sync it up. Bit of a nightmare as the diff is pretty wild.

jamesrhester:
This PR now replaced by #358

PR #302, opened 2022-09-16
add units for mass absorption coefficients

rowlesmr:
Enumerate the allowed units for mass absorption coefficients.

Needed by https://github.com/COMCIFS/Powder_Dictionary/pull/19


jamesrhester:
The layout error is a problem with the layout checking routine, I'll try and fix it but meanwhile this can be merged.

PR #305, opened 2022-10-04
Fix diffrn radiation

### LINKED ISSUE #168: cif_core.dic: DIFFRN_RADIATION is described as potentially looped, but is actually a set
The definition of the `DIFFRN_RADIATION` category states that:
```
Items may be looped to identify and assign weights to distinct wavelength
components from a polychromatic beam.
```

However, the category is formally described as a `Set` category.

The category description is also identical to the description of the `DIFFRN_RADIATION_WAVELENGTH` category and differs from the one provided in the `DDL1` version of the dictionary:
```
Data items in the DIFFRN_RADIATION category describe the radiation used in measuring
the diffraction intensities, its collimation and monochromatization before the sample.

Post-sample treatment of the beam is described by data items in
the DIFFRN_DETECTOR category.
```

There are about ~800 entries in the COD that contain looped instances of data items from the `DIFFRN_RADIATION` category. In most of these cases, the `diffrn_radiation_type` item is looped with other items from the  `DIFFRN_RADIATION_WAVELENGTH` category. For example, COD entry [2311995](https://www.crystallography.net/cod/2311995.html):
```
loop_
_diffrn_radiation_wavelength
_diffrn_radiation_wavelength_wt
_diffrn_radiation_type
_diffrn_radiation_wavelength_id
0.709237 1.000 K\a~1~ 1
0.713647 0.500 K\a~2~ 2
```

However, in a few select cases more than one item from the `DIFFRN_RADIATION` is looped, i.e.
COD entry [4002662](https://www.crystallography.net/cod/4002662.html):
```
loop_
_diffrn_radiation_type
_diffrn_radiation_probe
_diffrn_radiation_monochromator
_diffrn_radiation_filter_edge
_diffrn_radiation_xray_symbol
_diffrn_radiation_wavelength
_diffrn_radiation_wavelength_wt
'Cu K\a' x-ray 'Ni filter' 1.4879 'Cu K-L~3~' 1.540596 0.6681294
'Cu K\a' x-ray 'Ni filter' 1.4879 'Cu K-L~2~' 1.544493 0.3308458
```

As mentioned, there aren't that many instances of such loopness, however, the provided examples are from recent publications (years 2018, 2016). Maybe the `DIFFRN_RADIATION` category could be formally declared as looped and, in this way, allow items from the `DIFFRN_RADIATION_WAVELENGTH` child looped category to be stored in the same loop (as is done with the `ATOM_SITE` and `ATOM_SITE_ANISO` categories)?

jamesrhester:
This addresses and resolves #168. The cif_core dictionary management group had no objections to this course of action.

Closes #168.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -1611,9 +1611,9 @@ save_
 save_diffrn_radiation.type
 
     _definition.id                '_diffrn_radiation.type'
-    _alias.definition_id          '_diffrn_radiation_type'
     _definition_replaced.id       1
     _definition_replaced.by       '_diffrn_radiation_wavelength.type'
+    _alias.definition_id          '_diffrn_radiation_type'
     _definition.update            2022-10-04
     _description.text
 ;
@@ -1640,10 +1640,9 @@ save_
 save_diffrn_radiation.xray_symbol
 
     _definition.id                '_diffrn_radiation.xray_symbol'
-    _alias.definition_id          '_diffrn_radiation_xray_symbol'
     _definition_replaced.id       1
     _definition_replaced.by       '_diffrn_radiation_wavelength.xray_symbol'
-
+    _alias.definition_id          '_diffrn_radiation_xray_symbol'
     _definition.update            2022-10-04
     _description.text
 ;
@@ -1769,6 +1768,31 @@ save_diffrn_radiation_wavelength.id
 
 save_
 
+save_diffrn_radiation_wavelength.type
+
+    _definition.id                '_diffrn_radiation_wavelength.type'
+    _definition.update            2022-10-04
+    _description.text
+;
+    Details of the radiation source or energy spectrum.
+;
+    _name.category_id             diffrn_radiation_wavelength
+    _name.object_id               type
+    _type.purpose                 Describe
+    _type.source                  Recorded
+    _type.container               Single
+    _type.contents                Text
+
+    loop_
+      _description_example.case
+         'Mo K\a'
+         'Cu K\a'
+         'Cu K\a~1~'
+         'Cu K-L~2,3~'
+         'white-beam'
+
+save_
+
 save_diffrn_radiation_wavelength.value
 
     _definition.id                '_diffrn_radiation_wavelength.value'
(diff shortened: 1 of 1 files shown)


vaitkus:
The changes seem to properly correspond with the solution described in #168.

It was probably already been checked, but just in case, would any other items from the `DIFFRN_RADIATION` category should also be moved to `DIFFRN_RADIATION_WAVELENGTH`? The remaining items are:

- `_diffrn_radiation.collimation`.
- `_diffrn_radiation.filter_edge`.
- `_diffrn_radiation.inhomogeneity`.
- `_diffrn_radiation.monochromator`.
- `_diffrn_radiation.polarisn_norm`.
- `_diffrn_radiation.polarisn_ratio`.
- `_diffrn_radiation.probe`.

jamesrhester:
I did look at this previously and concluded that the remaining items in `diffrn_radiation` are likely to be constant for a particular setup. If the COD has instances of these items being looped and having different values for different rows of the loop that would be evidence that we should consider moving other data names to `diffrn_radiation_wavelength`, otherwise I think it best not to deprecate more than we have to.

vaitkus:
@jamesrhester thank you for clarifying. I checked the latest COD revision (r278520) and there indeed are no other looped data items from the `DIFFRN_RADIATION` category. However, if the decisions to deprecate items were mainly made on their prevalence in the COD, I must note that the looped `_diffrn_radiation_xray_symbol` data item with different values only appears in 3 entries, thus it might potentially be viewed as not significant enough to require changes in the dictionary (I am not advocating to revert the changes, just providing additional information).

I also read though the original issue again and noticed that it mentioned potential changes to the dREL code. Should the dREL code of items `_atom_type_scat.dispersion_imag` and `_atom_type_scat.dispersion_real` be changed in any way now that the `diffrn_radiation.type` is deprecated and the replacement `_diffrn_radiation_wavelength.type` is potentially looped?

jamesrhester:
Yes, that dREL code should be changed eventually, and in quite significant ways as the dispersion should be interpolated from tables rather than looked up for a particular X-ray tube line. That is definitely a separate issue to this one. For now the dREL is correct in the sense that it refers to a data name that is defined.

`xray_symbol` was included because logically it could be looped for separate components in a single beam.

vaitkus:
OK, I will file a separate issue for the dREL then. Other than that, everything looks great.

Issue #234, opened 2021-08-02
cif_core.dic: add the `_exptl_absorpt_special_details` data item


vaitkus:
In the 267887 revision of the COD there are over 1000 files that contain the `_exptl_absorpt_special_details` data item, although, no such item is defined in the `CIF_CORE` dictionary.

My initial assumption was that this data item was simply used instead of the properly defined `_exptl_absorpt_process_details` data item. However, some additional analysis revealed that in ~800 cases out of the 1000, both the `_exptl_absorpt_special_details` and the `_exptl_absorpt_process_details` data items are defined and often with differing values. For example, COD entry 7060029 [1] contains the following data item values:

- `_exptl_absorpt_process_details`:

```
<i>SADABS</i> (Krause et al., 2015)
```
- `_exptl_absorpt_special_details`:
```
In addition to absorption, multi-scan techniques also correct other slowly 
varying changes in scale factor over reciprocal space, such as those caused 
by x-ray beam inhomogeneity, goniometer imperfection, crystal decomposition, 
crystal longer than the x-ray beam cross section, absorption by the crystal 
mount etc.  These account, by and large, for much of the difference between 
estimated values of Tmin and Tmax and the _exptl_absorpt_correction_ values. 
```

I am unsure where the `_exptl_absorpt_special_details` tag originated from, however, I have noticed that all of the affected files seem to be generated by SHELXL. What is more, I found some SHELX-related material online [2] that provides a usage example of the `_exptl_absorpt_special_details` data item (see slide "Comments on sample absorption correction").

[1] http://www.crystallography.net/cod/7060029.cif@258905
[2] https://chemistry.harvard.edu/files/chemistry/files/tutorial_using_shelx_for_neutrons.pdf

jamesrhester:
I will pass this to the core DMG as clearly this is a data name that is useful, as shown by your ref [2].

jamesrhester:
The core DMG support the addition of this data name to the dictionary. Given its origin, we will need to include the all underscore version as an alias.

vaitkus:
Issue resolved by merging PR #234. Thanks, @rowlesmr.

PR #314, opened 2022-11-22
add the _exptl_absorpt.special_details data item

rowlesmr:
Addresses #234 

The examples given are a poor copy/paste from the examples of @vaitkus . Will probably need to be changed.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -9782,14 +9782,14 @@ save_exptl_absorpt.special_details
     loop_
       _description_example.case
 ;
-      X-ray beam inhomogeneity and goniometer imperfection account, by and large,
-      for much of the difference between estimated values of Tmin and Tmax and
-      the _exptl_absorpt_correction_ values.
+       X-ray beam inhomogeneity and goniometer imperfection account, by and
+       large, for much of the difference between estimated values of Tmin and
+       Tmax and the _exptl_absorpt_correction_ values.
 ;
 ;
-      X-ray linear absorption coefficient is wavelength dependent. The value of
-      0.420 (mm^-1^) shown in _exptl_absorpt_coefficient_mu is the sample
-      absorption coefficient for a wavelength of 1.5 \%A.
+       X-ray linear absorption coefficient is wavelength dependent. The value
+       of 0.420 (mm^-1^) shown in _exptl_absorpt_coefficient_mu is the sample
+       absorption coefficient for a wavelength of 1.5 \%A.
 ;
 
 save_
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.2.0
-    _dictionary.date              2022-10-04
+    _dictionary.date              2022-11-22
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.1.0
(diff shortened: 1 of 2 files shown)


cif_core.dic
```diff
@@ -26039,10 +26077,12 @@ save_
...
 ;
-         3.2.0                    2022-10-04
+         3.2.0                    2022-11-22
 ;
        Added data names to allow multi-data-block expression of data sets.
 
        Deprecated and replaced _diffrn_radiation.type and
        _diffrn_radiation.xray_symbol.
+
+       Added the _exptl_absorpt.special_details data item
```
vaitkus, line 26087:
```suggestion
       Added the _exptl_absorpt.special_details data item.
```

cif_core.dic
```diff
@@ -9756,6 +9756,44 @@ save_exptl_absorpt.process_details
...
 save_
 
+save_exptl_absorpt.special_details
+
+    _definition.id                '_exptl_absorpt.special_details'
+
+    loop_
+      _alias.definition_id
+         '_exptl_absorpt_special_details'
+         '_exptl.absorpt_special_details'
```
vaitkus, line 9766: The `_exptl.absorpt_special_details` alias should not be included as it was probably not used anywhere before. Aliases of other data items from this category that have the form `_exptl.<something>` (e.g. `_exptl.absorpt_coefficient_mu`) seem to have been taken from the mmCIF dictionary and are thus more meaningful. If needed, we can always add this alias in the future.

cif_core.dic
```diff
@@ -9756,6 +9756,44 @@ save_exptl_absorpt.process_details
...
+
+    _definition.update            2022-11-22
+    _description.text
+;
+    Details of the absorption correction process applied to the
+    measured intensities that cannot otherwise be given by other
+    _exptl_absorpt.* data names.
```
vaitkus, line 9773:
```suggestion
    measured intensities that cannot otherwise be given using
    other data items from the EXPTL_ABSORBT category.
```

cif_core.dic
```diff
@@ -9756,6 +9756,44 @@ save_exptl_absorpt.process_details
...
+;
+       X-ray beam inhomogeneity and goniometer imperfection account, by and
+       large, for much of the difference between estimated values of Tmin and
+       Tmax and the _exptl_absorpt_correction_ values.
+;
+;
+       X-ray linear absorption coefficient is wavelength dependent. The value
+       of 0.420 (mm^-1^) shown in _exptl_absorpt_coefficient_mu is the sample
+       absorption coefficient for a wavelength of 1.5 \%A.
+;
```
vaitkus, line 9793: The provided examples look good in my eyes, but in case "live" examples are preferred, I attach a file with the `_exptl_absorpt_special_details` data item values observed in the COD. Some are clearly not fit as examples, but there are definitely more than a few more meaningful ones. Each line in the file starts with a COD ID, followed by a tab and the data item value that was unfolded into a single line. In case you decide to go with any of these examples, I can easily provide references to the original CIF publication which could potentially be included into the example loop using the `_description_example.detail` data item (this seems to sometimes be done in examples, though, I am unsure if this is necessary in this case).

File with the values: [special-detail-values-from-the-COD.txt](https://github.com/COMCIFS/cif_core/files/10108594/special-detail-values-from-the-COD.txt)

jamesrhester: I think the examples currently provided are adequate to demonstrate ways in which this data name are used, so there is no need to for further examples. I think longer more detailed examples are useful when showing how a whole category, or several categories, interact.

cif_core.dic
jamesrhester: Agree.

cif_core.dic
rowlesmr: Mine are kind of already "live", inasmuch as they were heavily "inspired" by real ones. By "reference", do you mean to the journal paper where the data originally came from? I'll see what I can find.

cif_core.dic
rowlesmr: will fix.

Issue #278, opened 2021-11-05
Should measurand and associated SU data items always have the same content type


vaitkus:
I have noticed that several Measurand data items from the `CIF_CORE` dictionary have the `Integer` content type while the associated SU data items have the `Real` content type. Data items in question:
```
_diffrn_refln.counts_bg_1
_diffrn_refln.counts_bg_2
_diffrn_refln.counts_net
_diffrn_refln.counts_peak
_diffrn_refln.counts_total
```
Also note, that while it is possible to assign a real SU value to a integer measurand value using a separate data item, this does not seem to be possible using the alternative parenthesis notation. For example:
```
_value          43
_value_su       1
```
can be written as 43(1), but it is unclear how the same could be done for:
```
_value          43
_value_su       0.8
```
Two questions:
1. Should the previously mentioned `_diffr_refln.count_*` data items be changed in any way (e.g. by redefining SU items as integers)?
2. Can it be reasonably assumed that SU items and Measurand items will always have the same content type?

jamesrhester:
1. No, `_diffrn_refln.count_*` do not need to be redefined. I believe this behaviour is consistent with the SU of a Poisson-distributed value being the square root of the value.
2. No, as the example of Poisson counting statistics shows, it cannot be reasonably assumed.

The point that the bracket notation cannot handle some SU values for integers is worth making and we are fortunate that we now recognise that the SU can always be presented as a separate data value. I will add a comment on this to the DDLm chapter.

jamesrhester:
Here is the revised text of the relevant part of the draft DDLm chapter. Feel free to comment!

> DDLm provides a mechanism to specify
> that one data item holds the standard uncertainty (SU) of another data item.
> When using CIF formats, which allow the SU to be presented as part of a
> data value, this separate SU data name is assigned the SU value (see
> Table \ref{tab:DDLm-derive-CIF}). Where 
> both an appended SU and a separate SU data item are present for a data item, the
> values must agree. CIF users should also note that appending SU may
> not be possible in cases where an integer data type has an SU that is smaller
> than one (for example an SU of 0.8 on a value of 2), in which case the
> explicit SU data name is available.

vaitkus:
Thank you for the answers.

The revised text looks great, however, the revision might not actually be required after all. A more detailed explanation is provided below.

[...]

PR #315, opened 2022-11-22
add _diffrn_refln.intensity_* data names

rowlesmr:
Addresses #278 

Altered entries for  `_diffrn_refln.counts_bg_1`, `_diffrn_refln.counts_bg_2`, `_diffrn_refln.counts_net`, `_diffrn_refln.counts_peak`, and `_diffrn_refln.counts_total` and removed their `_su` counterparts

Added their `*.intensity_*` counterparts.

There were no existing definitions, and so I made up words which I think are in the correct vein. *They need to be changed!*

`_diffrn_refln.intensity_net` already existed, and so I updated some of it's details, but left the remainder intact.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -2014,27 +2014,26 @@ save_diffrn_refln.counts_bg_1
 
     _definition.id                '_diffrn_refln.counts_bg_1'
     _alias.definition_id          '_diffrn_refln_counts_bg_1'
-    _name.category_id             diffrn_refln
-    _name.object_id               counts_bg_1
-
-    _import.get                   [{'file':templ_attr.cif  'save':diffr_counts}]
-
-save_
-
-save_diffrn_refln.counts_bg_1_su
-
-    _definition.id                '_diffrn_refln.counts_bg_1_su'
-    _definition.update            2021-09-23
+    _definition.update            2022-11-22
     _description.text
 ;
-    Standard uncertainty of _diffrn_refln.counts_bg_1.
+    Background scattering 1 for the reflection.
+
+    Note that counts-per-second values should be converted to
+    total counts.
+
+    Standard uncertainties should not be quoted for these values.
+    If the standard uncertainties differ from the square root of
+    the number of counts, _diffrn_refln.intensity_* should be used.
 ;
     _name.category_id             diffrn_refln
-    _name.object_id               counts_bg_1_su
-    _name.linked_item_id          '_diffrn_refln.counts_bg_1'
-    _units.code                   none
-
-    _import.get                   [{'file':templ_attr.cif  'save':general_su}]
+    _name.object_id               counts_bg_1
+    _type.purpose                 Number
+    _type.source                  Recorded
+    _type.container               Single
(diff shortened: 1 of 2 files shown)


rowlesmr:
I think I've done all of that.

I have tweaked the definitions of `counts_peak` and `counts_total` to say that there is a region of interest that includes only the peak, or includes the peak and some background, respectively. I think that should cover area detectors.

The definition of `_diffrn_refln.intensity_net` currently refers to 'net intensity' from the 'diffraction counts' after applying attenuator and standard scales. Are the 'diffraction counts' bkg-corrected? should we explicitly mention that you are correcting the value given in `_diffrn_refln.counts_net`?

jamesrhester:
Almost perfect. I would still make `_diffrn_refln.counts_net` a `Real` value because for area detector background subtraction it is almost certain that a 2D sloped background will have been applied resulting in real numbers.  If despite this we leave this quantity as an integer we are faced with the original difficulty of potentially not being able to write the attached su to sufficient accuracy for counts of 1/2/3.

As an alternative we can create a new dataname `_diffrn_refln.uncorrected_intensity_net` to capture non-integral net counts before standard + attentuator scaling.

Furthermore, an almost absent peak might statistically result in a value less than zero, so the lower limit probably should be dropped.

rowlesmr:
> Almost perfect. I would still make `_diffrn_refln.counts_net` a `Real` value because for area detector background subtraction it is almost certain that a 2D sloped background will have been applied resulting in real numbers. If despite this we leave this quantity as an integer we are faced with the original difficulty of potentially not being able to write the attached su to sufficient accuracy for counts of 1/2/3.
>
> Furthermore, an almost absent peak might statistically result in a value less than zero, so the lower limit probably should be dropped.

This could also apply with point detectors. Changed. I've left the units as `counts`, as they are derived directly from count values.

cif_core.dic
```diff
@@ -2014,81 +2014,97 @@ save_diffrn_refln.counts_bg_1
...
+;
+    Counts measured in the reflection peak after background
+    subtraction.
+
+    Note that counts-per-second values should be converted to
+    total counts.
+
+    The value for this data name must be derived from count values.
+    If not, _diffrn_refln.intensity_* should be used.
+;
```
jamesrhester, line 2085: If I measure counts in the peak P for 3 times longer than the background B, then I could calculate `counts_net` as P-3B, or I could calculate as P/3 - B. Both are "derived from count values" and both are correct.  If we want to use this for area detector single-crystal data, then our final value could be more complicated, but still only involve counts and relative scaling factors.

For avoidance of the above ambiguity I think we need to be more specific and state that background counts should be scaled to the peak counts before subtraction. This is the usual practice so I don't think we are changing anything here.

rowlesmr: I hadn't thought that someone _wouldn't_ scale bkg to peak first! I'll add that.

jamesrhester:
So let's wait to see what @vaitkus thinks before merging. I'm happy with this.

PR #321, opened 2023-01-10
Define "integer" to include zero.

### LINKED ISSUE #280: ddl.dic: `_type.contents` definition of an Integer excludes 0
The `_type.contents` data item describes the Integer type as:
```
positive or negative integer number
```
Since zero is neither positive nor negative (or both positive and negative), it is currently excluded from the allowed values.

rowlesmr:
Will close #280 

Previously, zero was excluded.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -10,7 +10,7 @@ data_DDL_DIC
     _dictionary.title             DDL_DIC
     _dictionary.class             Reference
     _dictionary.version           4.1.0
-    _dictionary.date              2023-01-09
+    _dictionary.date              2023-01-11
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/ddl.dic
     _dictionary.ddl_conformance   4.1.0
@@ -1630,7 +1630,7 @@ save_type.contents
 
     _definition.id                '_type.contents'
     _definition.class             Attribute
-    _definition.update            2023-01-09
+    _definition.update            2023-01-11
     _description.text
 ;
     Syntax of the value elements within the container type. Where the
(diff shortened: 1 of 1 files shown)


ddl.dic
```diff
@@ -1729,7 +1729,7 @@ save_type.contents
 ;
          Integer
 ;
-         positive or negative integer number
+         An integer. Positive and negative integers, and zero, are included.
```
vaitkus, line 1732:
```suggestion
         An integer. Belongs to a numerical set that consists of positive integers, negative integers and the number zero.
```

vaitkus: Simply proposed an alternative formulation of the same statement. Feel free to ignore if it seems too verbose.

rowlesmr: It does seem a little self-referential to define a thing in terms of itself.

[Wolfram|Alpha](https://www.wolframalpha.com/examples/mathematics/numbers/integers) define integers as:
> The set of integers consists of the counting numbers, their negations and zero.

Perhaps we could use:
> An integer. Integers consist of the positive and negative counting numbers, and zero.

vaitkus: The term "negative counting numbers" looks a bit off to me as counting numbers by definition should not be negative. Maybe something like "`A number from the set of all integers.`" would work? We could potentially further specify what makes up the integer set, but this seems like a well established mathematical term so there is probably no need to be too verbose. The important thing in my eyes is to specify two things:
- That the allowed values include a zero.
- That the `Integer` in question refers to the mathematical concept of an integer and not to a programmers view of an integer.

Both things seem to be covered by the phrase "set of all integers". What do you think?

rowlesmr: That sounds good to me.

"`A number from the set of all integers.`"

vaitkus (approved):

PR #322, opened 2023-01-10
formatted _enumeration_set.detail values to be sentences.

rowlesmr:
Sentences now start with a capital letter, end with a full stop. Previously, they were a mixture.


vaitkus:
I guess that in this case the `_definition.update` can be left unmodified since this is most likely a one-time cosmetic change of this scale. Thank you for the changes.

Issue #189, opened 2020-10-28
DDLm definition of the _exptl_crystal.colour data item is not compatible with the DDL1 definition


vaitkus:
The  `_exptl_crystal.colour` data item in the `DDLm` version of the `CIF_CORE` dictionary is defined as a list of 3 elements ([`_exptl_crystal_appearance.general`, `_exptl_crystal_appearance.intensity`, `_exptl_crystal_appearance.hue`]) while the `DDL1` version of the  `CIF_CORE` dictionary defines it as a simple concatenated text string without a container. Since neither `CIF_1.1` nor `DDL1` support lists or other container types, CIF files that contain the `_exptl_crystal.colour` data item simply cannot be simultaneously valid according to both  the `DDL1` and the `DDLm`.  It should be noted that the  `_exptl_crystal_colour` (`_exptl_crystal.colour`) data item has been widely used and appears in at least ~78% of the COD entries (r249495, ~450 000 total entries).

As discussed in PR #188, one possible solution could be to revert the `_exptl_crystal.colour` data type back to the original DDL1 type (text string with no container) and define an entirely new data item for the structured (list) values.

jamesrhester:
This is indeed a serious problem. We must fulfill our promise that no changes are made compared to DDL1. Therefore we should simply fix the definition of this data name to be equivalent to DDL1. The dREL can construct the DDL1 string using the values in `exptl_crystal_appearance`. It does not seem necessary to create a list-valued version of this data name as it is not referenced anywhere else.

vaitkus:
Issue was resolved by merging PR #325

PR #325, opened 2023-01-12
Change the _exptl_crystal.colour data item from a list to a text string

vaitkus:
This PR addresses issue #189.

While drafting this PR I made a few assumptions that should be addressed before merging:
- I introduced quite a strict format for the color text string based on the values of items from the `EXPTL_CRYSTAL_APPEARANCE` category to mimic the format of the removed list data structure. However, alternatively we might want to remove any formatting requirements on the field so it could also be used as a catch-all field (read `special_details`) for crystal colour data. There are a few instances in the COD of entries which seemingly cannot be described by the items from the `EXPTL_CRYSTAL_APPPEARANCE` category (e.g. `dichroic dark purple/pale blue`), but they are not very numerous. Should I remove the formatting requirements (the dREL part can still remain as a fallback mechanism)? 
- I added a statement about `EXPTL_CRYSTAL_APPEARANCE` items being preferred since modular structured data is usually a better choice.
- The dREL code that I added should work well in the most general case when all items from the `EXPTL_CRYSTAL_APPEARANCE` category are provided, but would most likely produce a strangely formatted string when at least one of the items is missing/undefined (string would contain leading, trailing or double spaces). This could be rewritten to check if the item actually has a value and only then concatenate it to the string, however, I am not sure if there is an operator for such a check in dREL. Any suggestions?

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10038,12 +10038,12 @@ save_exptl_crystal.colour
          'red'
          'colourless'
 
-    _method.purpose               Evaluation
+    _method.purpose               Definition
     _method.expression
 ;
     With c as exptl_crystal_appearance
 
-     _exptl_crystal.colour = c.general + " " + c.intensity + " " + c.hue
+     _enumeration.default = c.general + " " + c.intensity + " " + c.hue
 ;
 
 save_
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10000,10 +10000,7 @@ save_exptl_crystal.colour
     _definition.update            2023-01-12
     _description.text
 ;
-    Colour description of the crystal expressed as a combination of
-    the _exptl_crystal_appearance.general,
-    _exptl_crystal_appearance.intensity and _exptl_crystal_appearance.hue
-    data item values separated by sequences of a single space symbol.
+    Colour description of the crystal.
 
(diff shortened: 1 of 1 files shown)


jamesrhester:
I've altered the dREL to compute a default value. I don't think extra whitespace is an issue in the dREL, but if you do want to check for missing values you can use `Is_missing(arg)`. This is the only way to check. You could also post-process the string to remove trailing and sequential whitespace but dREL string processing is pretty basic.

vaitkus:
@jamesrhester thank you for the changes. I further updated the dREL code based on your comments.

I will repeat one of my previous questions for visibility:
* I introduced quite a strict format for the color text string based on the values of items from the `EXPTL_CRYSTAL_APPEARANCE` category to mimic the format of the removed list data structure. However, alternatively we might want to remove any formatting requirements on the field so it could also be used as a catch-all field (read `special_details`) for crystal colour data. There are a few instances in the COD of entries which seemingly cannot be described by the items from the `EXPTL_CRYSTAL_APPPEARANCE` category (e.g. "dichroic dark purple/pale blue"), but they are not very numerous. Should I remove the formatting requirements (the dREL part can still remain as a fallback mechanism)?

Issue #205, opened 2021-03-18
Unit of measurement of the _refine_diff.density_* data items


vaitkus:
Units of measurement of the `_refine_diff.density_min`, `_refine_diff.density_min_su`, `_refine_diff.density_max`, `_refine_diff.density_max_su`, `_refine_diff.density_rms` and `_refine_diff.density_rms_su` data items are currently assigned using the following definition method:
```
     If (_diffrn_radiation.probe == "neutron")  _units.code =  "femtometres"
Else If (_diffrn_radiation.probe == "electron") _units.code =  "volts"
Else                                            _units.code =  "electrons"
```
Based on the DDL1 version of the `CIF_CORE` dictionary the `electrons` unit should most likely be `electrons_per_angstrom_cubed`, however, I am unsure if the same can be extrapolated to `femtometres` and `volts`. For example, `femtometres_per_angstrom_cubed` would look extremely strange as a unit of density given that both the `femtometers` and `angstroms` are units of length. What would be the proper units (if any) for the rest of the radiation probes?

jamesrhester:
X-rays and electrons both produce electrons per angstrom cubed. Neutron scattering uses "scattering length" instead of "electrons". So the density becomes scattering length in femtometres per cubic angstrom, and the lengths on the top and bottom can cancel out to get "angstrom^-2". Typically (e.g.  https://www.ncnr.nist.gov/resources/activation/) the units are then 10^-6/A^2, which is just the same as femtometres/cubic angstrom. So I think it is OK to have femtometres/cubic angstrom as this makes the 10^-6 scaling explicit.

rowlesmr:
> What units should be applied to gamma radiation (`gamma` is one of enumeration values given for `_diffrn_radiation.probe`)?

I would guess that `gamma` would be the same units as `x-ray`; they're both photons.

jamesrhester:
Yes that new dREL looks generally OK, how about a PR?

vaitkus:
Great, thanks. I created the PR, just wanted to make sure first that the dREL is OK to avoid tweaking it in 6 different places.

PR #326, opened 2023-01-12
Fix scattering density units in dREL

vaitkus:
This PR resolves issue  #205.

Note that I also slightly changed the phrasing in the description of the `REFINE_DIFF` category. Specifically, I changed the phrase "electron density limits" to "scattering density limits" since the data items in the category can describe both electron densities and scattering length density. @jamesrhester , @rowlesmr  is the term `scattering density limits` appropriate or is there a better one?

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.2.0
-    _dictionary.date              2023-01-10
+    _dictionary.date              2023-01-12
     _dictionary.uri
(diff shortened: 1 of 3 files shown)


rowlesmr:
"scattering density limits" is probably OK. I found [this paper](http://dx.doi.org/10.2138/am-2003-5-607)) which plotted their neutron difference fourier maps in "scattering units". [This SI](https://journals.iucr.org/b/issues/2005/06/00/bs5020/bs5020sup7.pdf) uses the phrase "nuclear scattering density" for their neutron Fourier maps. [Jeremy et al.](http://pd.chem.ucl.ac.uk/pdnn/refine2/fourier.htm) says "scattering density".

PR #328, opened 2023-01-13
Fix minor typos

vaitkus:
This PR fixes a few minor formatting and syntax issues. @jamesrhester, @rowlesmr a review from a native speaker would be highly appreciated.


rowlesmr:
Those all look fine to me. I don't know if any warrant changing the update date, but strictly, it should be updated. There will be more typos to fix!

.

Related question, maybe should be it's own issue: is there guidance on the number of characters on a line of a description text? They all start at column 5 (as set by formatting rules), but many seem to finish by column ~70, which is quite short of the allowed 80.

rowlesmr:
@vaitkus I have some _description_example.detail formatting I'd like to add to this PR. I can't edit this PR; are you able to allow editing on it? Otherwise I can submit my own PR.

vaitkus:
@rowlesmr, thank you for the review. Editing on the branch is enabled, but I think you do not have push permission on the main repository. I cannot change this since I am only a collaborator on this project and thus do not have access to more sensitive settings. So probably a PR on this branch (or the main branch) is the best currently available approach.

> Related question, maybe should be it's own issue: is there guidance on the number of characters on a line of a description text? They all start at column 5 (as set by formatting rules), but many seem to finish by column ~70, which is quite short of the allowed 80.

Such things are generally covered in the dictionary style guide: https://github.com/COMCIFS/comcifs.github.io/blob/master/accepted/ddlm_dictionary_style_guide.md. According to section 1 and 2.1 of the guide, there is a hard limit of 80 characters per line (with some exceptions), but I do not think that it is explicitly required to always reach this limit. If you want to discuss this topic further, it would be better to open a separate issue, probably in the comcifs.github.io repo.

rowlesmr:
> @rowlesmr, thank you for the review. Editing on the branch is enabled, but I think you do not have push permission on the main repository. I cannot change this since I am only a collaborator on this project and thus do not have access to more sensitive settings. So probably a PR on this branch (or the main branch) is the best currently available approach.

Yeah, rereading the rules seems to say I need to have write access to write to existing PRs. I'll do my own branch. Don't wait for me.

PR #336, opened 2023-01-16
Explicitly add _diffrn_source.details to the list items that replace _diffrn_source.description

vaitkus:
This PR explicitly add `_diffrn_source.details` to the list of items that replace `_diffrn_source.description` since this item was already mentioned in the human-readable part of the definition.


vaitkus:
Is this PR OK to merge?

PR #337, opened 2023-01-16
Reword definition of _exptl_crystal.size_mid to match that provided in mmCIF

vaitkus:
Not sure if this is significant, but the definitions of `_exptl_crystal.size_mid` differ slightly in CIF_CORE and mmCIF [1]. The difference is in a single word ("medial" vs "median"), both options of which seem to mean the same thing (to my level of understanding). Feel free to reject the PR if the change is indeed only nominal.

[1] https://mmcif.wwpdb.org/dictionaries/ascii/mmcif_std.dic

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.2.0
-    _dictionary.date              2023-01-13
+    _dictionary.date              2023-01-16
     _dictionary.uri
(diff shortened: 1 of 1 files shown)


jamesrhester:
"Medial" is not a familiar word for me. According to https://en.wiktionary.org/wiki/medial it pertains to the mean or average, which is a distinct meaning from "median". So changing this word is significant. Vol G refers to the "medial" dimension also for the DDL1 equivalent, so we should prefer "medial". I suspect that this word was chosen to be less specific than "median" or "mean", both of which would imply both a distinct number of dimensions that had been counted, and that at least half of them had been determined, whereas "medial" suggests a more approximate estimate of the general dimensions.

PR #351, opened 2023-02-03
Remove outdated dREL and update the content type in several definitions

### LINKED ISSUE #346: Contradictory definition of _citation_editor.id
The `_citation_editor.id` is defined as having a `Single` container with an `Implied` content type. However, the evaluation dREL of this item defines it as an Array of two elements:
```
    _citation_editor.id = [_citation_editor.citation_id,_citation_editor.ordinal]
```
Note, that the two items used to populate the array have different content types (`Word`, `Integer`). This seems like a bit of a deviation from current DDLm/dREL practices employed in CIF_CORE.

The `_citation_editor.id` data item seems to currently only be used as a convenient way of aggregating values since it is not explicitly specified as the primary key of the CITATION_EDITOR category (the key is currently a composite one and consists of the `_citation_editor.citation_id`, `_citation_editor.ordinal` pair).

One quick way to resolve the inconsistency could be to update the dREL code to concatenate the two values instead of putting them in an array, e.g.:
```
    _citation_editor.id = _citation_editor.citation_id + "_" + _citation_editor.ordinal
```
The underscore symbol was used as a connector to ensure that unique key values can always be generated (e.g. that concatenation of values 1,55 and 15,1 produce different strings).

Alternatively, the id item could be redefined to have a completely arbitrary value and potentially even become the primary category key.

What approach should be taken?

vaitkus:
Resolves #346.

This PR removes the outdated dREL from `_citation_author.key` and `_citation_editor.id`.

Note, that these items are still not explicitly specified as category keys in the definitions of  corresponding categories, but this should be addressed in a separate PR if needed.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -16186,23 +16186,6 @@ save_citation_author.citation_id
 
 save_
 
-save_citation_author.key
-
-    _definition.id                '_citation_author.key'
-    _definition.update            2023-02-03
-    _description.text
-;
-    Value is a unique key to a set of CITATION_AUTHOR items in a looped list.
-;
-    _name.category_id             citation_author
-    _name.object_id               key
-    _type.purpose                 Key
-    _type.source                  Related
-    _type.container               Single
-    _type.contents                Word
-
-save_
-
 save_citation_author.name
 
     _definition.id                '_citation_author.name'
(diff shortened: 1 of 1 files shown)


jamesrhester:
I have removed both these data names as they were created purely as a way to include both category keys in a single-value data name. Happy to merge if everyone else is happy with this change.

vaitkus:
No objection from me.

PR #360, opened 2023-03-27
Update the Current_row() function call.

vaitkus:
This PR implements the changes to the Current_Row() function discussed in https://github.com/COMCIFS/dREL/issues/6.

Note, that it should only be merged after https://github.com/COMCIFS/dREL/pull/9 is merged.


jamesrhester:
I made this change in passing while preparing the latest change to #359 so this PR is no longer needed.

vaitkus:
Great, thanks. Will take a look there.

PR #366, opened 2023-04-04
Add a dREL definition method for the _citation.id data item

### LINKED ISSUE #363: Introduce a default value to the `_citation.id` data item
During the discussion in https://github.com/COMCIFS/cif_core/issues/58 it was decided to remove almost all default enumeration values due to them being potentially misleading. However, defaults related to `PUBL` were left untouched on the basis that they do not relate to data as much (hopefully the same logic extends to the  `_citation.id` data item).

The  `_citation.id` data item serves as a primary key of the `CITATION` category. However, due to legacy reasons the item is assigned the `Encode` purpose rather than the `Key` purpose which is assigned to keys with arbitrary values. Specifically, the `Encode` purpose is needed since `_citation.id`  has one special value `primary` which indicates that the citation is the primary one. As a result, during the validation process the missing `_citation.id`  data item is reported even if there is only a single packet in the `CITATION` loop (a very common occurrence), since the value cannot be automatically elided.

I propose to resolve this issue by introducing `primary` as the default. This will work well with loops that only have a single entry (which is most likely the primary) and the validation will still properly detect the missing key in cases when there is more than one loop since the default key will not be unique.

vaitkus:
Closes #363.

Note, that I am not completely sure if the `len(citation)` can be used to get the category packet count.


vaitkus:
The `len()` built-in is defined for strings and lists (https://github.com/COMCIFS/dREL/blob/73272db21637ce8de61bbb2bd41652d40ae392c9/builtins.md?plain=1#L68), but you are completely correct that its application on a categories is ambiguous. The propose `has_single_row(<cat>)` functions seems ok, but how about having a more generic version like `Get_row_count(<cat>)`?

rowlesmr:
Numpy has `shape`, which is a property of an array. It is a tuple of the lengths of all the dimension. Could you make that?

```
shape(category) = [ 23 3 ]
```
where there are 23 rows and 3 columns
?

rowlesmr:
also, if the need ever arises, you could make `width(category)` be a thing.

rowlesmr:
@jamesrhester which CIF release version is this change going into?

The change description at the bottom is in v3.2.0, abut we're now in v3.3.0.

jamesrhester:
If it's not in 3.2 by now it is in 3.3. I'll update accordingly.

jamesrhester:
It will be 3.3 if it is not in 3.2 already.

PR #368, opened 2023-04-09
Add ability to record an arbitrary number of form-factor coefficients

### LINKED ISSUE #357: 11 coefficient scattering factors? Cromer-Mann vs Waasmaier & Kirfel
Are there any particular reasons that the 11-coefficient scattering factors [1] aren't included in core?

[1]    D. Waasmaier & A. Kirfel, New Analytical Scattering Factor Functions for Free Atoms and Ions for Free Atoms and Ions, Acta Cryst. (1995). A51, 416-413

rowlesmr:
Will close #357, eventually.

I've added the ability to record an arbitrary number of coefficients for the conventional sum-of-Gaussians (a la Cromer-Mann and Wassimier-Kirfel), and the exponential polynomial model (cf Fox), as well as an inverse Mott-Bethe relationship parameterised as a sum of Gaussians.

Each of the models also has an upper/lower limit, so that an element can have multiple models associated with it for different s ranges.

If this looks OK, I'll have a go at looking at 
>The tables in `templ_enum.cif` would then have a list of numbers for each atom type and we would change the dREL for `_refln.form_factor_table` to access `_atom_type_scat.Gaussian_coeff`.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.2.0
-    _dictionary.date              2023-04-04
+    _dictionary.date              2023-04-09
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.1.0
@@ -27440,7 +27440,7 @@ save_
        Changed the content type of the _journal.paper_doi data item from
        'Code' to 'Text' and added an example case.
 ;
-         3.2.0                    2023-04-04
+         3.2.0                    2023-04-09
 ;
        Added data names to allow multi-data-block expression of data sets.
 
@@ -27514,4 +27514,8 @@ save_
 
        Changed the purpose of the _diffrn_radiation_wavelength.id data item
        from 'Encode' to 'Key'.
+
+       Added _atom_type_scat.exponential_polynomial_coefs, Gaussian_coefs, and
+       inv_Mott-Bethe_coefs to allow for an arbitrary number of coefficients in
+       the definition of form factors.
 ;
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -24801,13 +24801,13 @@ save_atom_type_scat.hi_ang_fox_coeffs
 
 save_
 
-save_atom_type_scat.inv_Mott-Bethe_coefs
+save_atom_type_scat.inv_Mott_Bethe_coefs
 
-    _definition.id                '_atom_type_scat.inv_Mott-Bethe_coefs'
+    _definition.id                '_atom_type_scat.inv_Mott_Bethe_coefs'
     _definition.update            2023-04-09
     _description.text
 ;
-    The set of Gaussian coefficients for use in the inverse Mott-Bethe
+    The set of Gaussian coefficients for use in the inverse Mott_Bethe
     relationship for generating X-ray scattering factors:
     [ c_1, d_1, c_2, d_2, ... , c_N, d_N, e ].
 
@@ -24818,7 +24818,7 @@ save_atom_type_scat.inv_Mott-Bethe_coefs
     number.
 ;
     _name.category_id             atom_type_scat
-    _name.object_id               inv_Mott-Bethe_coefs
+    _name.object_id               inv_Mott_Bethe_coefs
     _type.purpose                 Measurand
     _type.source                  Derived
     _type.container               Matrix
@@ -24828,17 +24828,17 @@ save_atom_type_scat.inv_Mott-Bethe_coefs
 
 save_
 
-save_atom_type_scat.inv_Mott-Bethe_coefs_su
+save_atom_type_scat.inv_Mott_Bethe_coefs_su
 
-    _definition.id                '_atom_type_scat.inv_Mott-Bethe_coefs_su'
+    _definition.id                '_atom_type_scat.inv_Mott_Bethe_coefs_su'
     _definition.update            2023-04-09
     _description.text
 ;
-    Standard uncertainty of _atom_type_scat.inv_Mott-Bethe_coefs.
+    Standard uncertainty of _atom_type_scat.inv_Mott_Bethe_coefs.
 ;
     _name.category_id             atom_type_scat
-    _name.object_id               inv_Mott-Bethe_coefs_su
-    _name.linked_item_id          '_atom_type_scat.inv_Mott-Bethe_coefs'
+    _name.object_id               inv_Mott_Bethe_coefs_su
+    _name.linked_item_id          '_atom_type_scat.inv_Mott_Bethe_coefs'
     _type.purpose                 SU
     _type.source                  Derived
     _type.container               Matrix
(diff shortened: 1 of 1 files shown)


rowlesmr:
This only allows the use of `List`s to record coefficients. Eg

```
loop_
_atom_type_scat.symbol
_atom_type_scat.Gaussian_coefs
_atom_type_scat.Gaussian_lower_limit
_atom_type_scat.Gaussian_upper_limit
_atom_type_scat.exponential_polynomial_coefs
_atom_type_scat.exponential_polynomial_lower_limit
_atom_type_scat.exponential_polynomial_upper_limit
Al [ 1.1151 6.420  3.0387 1.9002  0.7426 1.5936 31.5472 1.9646 85.0886 ] 0 2 [ 0.67975 -0.08756 -0.95431  0.72294 ] 2 6
S  [ 0.8669 6.9053 1.4679 5.2034 22.2151 1.4379  0.2536 1.5863 56.172  ] 0 2 [ 1.10400 -0.40325  0.20094 -0.26058 ] 2 6
```

To allow an arbitrary number of coefficients _and_ the ability to have a tag for the different coefficients is a little tricky, but needs a fair bit more infrastructure

```
loop_
_atom_type_scat.symbol
_atom_type_scat.order
_atom_type_scat.Gaussian_a
_atom_type_scat.Gaussian_b
_atom_type_scat.exponential_polynomial_a
Al 0 1.1151  0       0.67975
Al 1 6.420   3.0387 -0.08756
Al 2 1.9002  0.7426 -0.95431
Al 3 1.5936 31.5472  0.72294
Al 4 1.9646 85.0886  .
S  0 0.8669  0       1.10400
S  1 6.9053  1.4679 -0.40325 
S  2 5.2034 22.2151  0.20094 
S  3 1.4379  0.2536 -0.26058 
S  4 1.5863 56.172   .

loop_
_atom_type_scat_limits.symbol
_atom_type_scat_limits.Gaussian_lower
_atom_type_scat_limits.Gaussian_upper
_atom_type_scat_limits.exponential_polynomial_lower
_atom_type_scat_limits.exponential_polynomial_upper
Al 0 2 2 6
S  0 2 2 6
```

cif_core.dic
```diff
@@ -24628,6 +24801,94 @@ save_atom_type_scat.hi_ang_fox_coeffs
 
 save_
 
+save_atom_type_scat.inv_Mott_Bethe_coefs
+
+    _definition.id                '_atom_type_scat.inv_Mott_Bethe_coefs'
+    _definition.update            2023-04-09
+    _description.text
+;
+    The set of Gaussian coefficients for use in the inverse Mott_Bethe
```
vaitkus, line 24810:
```suggestion
    The set of Gaussian coefficients for use in the inverse Mott-Bethe
```

cif_core.dic
```diff
@@ -24542,6 +24542,179 @@ save_atom_type_scat.dispersion_source
...
+
+    f(s; Z) = exp(Sum( a~i~ * s^i^), i=0:N))
+
+    where s = sin(theta)/lambda.
+;
+    _name.category_id             atom_type_scat
+    _name.object_id               exponential_polynomial_coefs
+    _type.purpose                 Measurand
+    _type.source                  Derived
+    _type.container               Matrix
```
vaitkus, line 24562: Maybe `List` would be more suitable here?

This comment applies to other places in this PR that use the `Matrix` types as well.

rowlesmr: Yes. It does.

rowlesmr:
Yes. There should be a unique set of coefficients for any atom at any s.

vaitkus:
Is the `_atom_type_scat.exponential_polynomial_lower_limit` intended to always be inclusive and `_atom_type_scat.exponential_polynomial_upper_limit` always exclusive? Is yes, then it would make sense to explicitly specify this in the corresponding definitions.

Alternatively, since `*_lower_limit`/`*_upper_limit` describe a continuous range, maybe they could be rewritten as a single item using the `Range` content type (since we are not worried about compatibility with DDL1)? However, under the current definition of the this content type it would also be mean that the range is inclusive on both ends. I raised an issue with a suggestion to augment the current notation (see #372), but it is unclear when (and if) it is going to be approved.

rowlesmr:
Lower limits are always inclusive, upper are always exclusive. The definitions already say that.

It could be a range (it would be the first in the dictionary), if it were possible to have ex/inclusive ranges.

vaitkus:
> Lower limits are always inclusive, upper are always exclusive. The definitions already say that.

I apologise, I missed that part on the initial read and did not go back to check. With this in mind, I guess there is no real need to force it into a range type (as I initially suggested).

jamesrhester:
(Copied from #372 ) I'm not sure I understand the need for exclusive ranges here. An exclusive limit on the range of applicability means that any point arbitrarily close to the limit is still valid, and from continuity arguments we can see that the value of the first applicable point from the next range should be identical to the last value calculated from the previous range. So I'm not seeing what the point of an exclusive limit is in this case.

vaitkus:
Technically, ranges `[a, b]` and `[b, c]` overlap on a single value `b` whereas `[a, b)` and `[b, c)` do not.

However, I am unsure if this could give rise to any serious ambiguity.

rowlesmr:
> The question is, in a file with multiple alternative scattering factor coefficients tabulated, how is the set that was used for `F_calc` calculations indicated?

I would say that the scattering factors defined in the block containing the atom_type are the ones to use; from what I understand, that is the way it currently works, with default values being drawn from `_atom_type.symbol`.

```
data_str1
_blah blah

loop_
	_atom_site.label
	_atom_site.type_symbol
	_atom_site.fract_x
	_atom_site.fract_y
	_atom_site.fract_z
Al1 Al3+  0 0 0.148
O1 O2-  0.306 0 0.25

loop_
_atom_type_scat.symbol
_atom_type_scat.Gaussian_coefs
_atom_type_scat.Gaussian_lower_limit
_atom_type_scat.Gaussian_upper_limit
Al3+ [ 1.1151 6.420  3.0387 1.9002  0.7426 1.5936 31.5472 1.9646 85.0886 ] 0 2
O2-  [ 0.8669 6.9053 1.4679 5.2034 22.2151 1.4379  0.2536 1.5863 56.172  ] 0 2


data_str2
_blah blah

loop_
	_atom_site.label
	_atom_site.type_symbol
	_atom_site.fract_x
	_atom_site.fract_y
	_atom_site.fract_z
Sr Sr  0 0 0.148
O O2-  0.306 0 0.25

loop_
_atom_type_scat.symbol
_atom_type_scat.Gaussian_coefs
_atom_type_scat.Gaussian_lower_limit
_atom_type_scat.Gaussian_upper_limit
Sr [ 1.1151 6.420  3.0387 1.9002  0.7426 1.5936 31.5472 1.9646 85.0886 ] 0 2 
O2-  [ 1 2 3 4 5 6 7 8 ] 0 2
```

jamesrhester:
If you are asking about the situation where information for an experiment is distributed over multiple data blocks, as we are doing in powder CIF, then scattering coefficients should only be given once, or else agree completely, or else if you really want different values for O2- then you have to assign it a different atomic type. The concept here is that the form factor is a property of the atom type, so can't have different values for the same atom type. Can you conceive of a situation where different form factors would be used for the same atom type?

rowlesmr:
Then I don't understand your question. In a file with multiple alternative scattering factor coefficients tabulated, if they must agree, then why would there be a need to indicate which set was used for F_calc calculations?

jamesrhester:
> Then I don't understand your question. In a file with multiple alternative scattering factor coefficients tabulated, if they must agree, then why would there be a need to indicate which set was used for F_calc calculations?

OK, I've reviewed the thread and I think I see where we've misunderstood each other. My [original question](https://github.com/COMCIFS/cif_core/pull/368#issuecomment-1502727494) was premised on the misunderstanding that different parameterisations for the same atom and s would be possible. @rowlesmr subsequently [replied](https://github.com/COMCIFS/cif_core/pull/368#issuecomment-1504685563) that these would be unique for a given atom and s, which I am happy with. My most recent answer was based on me forgetting that this was my question originally, and essentially supplying @rowlesmr's answer back to him in the context of multiple data blocks. So I think we are in agreement, sorry if my wires got crossed.

> Can we differentiate between data present in the CIF and data made up through dREL?

dREL provides an algorithm for deriving values of data names. It is up to the user of the dictionary (where a "user" may be software) to decide how best to use that information. They could use it for validation, to fill in any missing rows, or to completely replace contents, whatever, their choice. In other words, the user of the dictionary knows what was in the CIF file, and can use that knowledge however they wish. What we as dictionary writers need to ensure is that the algorithm gives correct answers.  Not sure if that covers your question.

rowlesmr:
Roger. So it seems that the application of the dREL is implementation-defined.  I think I can work with that. I'll try and make something up.

PR #371, opened 2023-04-10
Create `ATOM_SCAT_TABLE` category

### LINKED ISSUE #216: cif_core.dic: description of the _atom_type_scat.versus_stol_list is misleading
The current description of the `_atom_type_scat.versus_stol_list` data item is as follows:
```
    List of scattering factors as a function of sin theta on lambda.
    List has the form [[<stol value 1> <scatfac 1>] [<stol value 2> <scatfac 2>] ....]
    in increments of 0.01, increasing from 0.0.
```
However,  after changes made in commit 0bc98b3 the described form of `[[<stol value 1> <scatfac 1>] [<stol value 2> <scatfac 2>] ... ]` does not match the rest of the definition that defines the value as a non-nested array of real numbers:
```
    _type.container               Array
    _type.dimension               '[]'
    _type.contents                Real
```

rowlesmr:
will close #216

eventually.

Starting with reverting `_atom_type_scat.versus_stol_list` to the DDL1 defintion.


cif_core.dic
```diff
@@ -27546,17 +27667,27 @@ save_
...
+
 ;
          3.3.0                    2023-06-16
 ;
        # Please update the date above and describe the change below until
        # ready for the next release
 
+       Rewrote the _atom_type.radius_contact evaluation method as a definition
+       method.
+
```
vaitkus, line 27685:
```suggestion
```

vaitkus: Seems like an artifact introduced during the last sync with the `master` branch. Should be removed.

cif_core.dic
```diff
@@ -27546,17 +27667,27 @@ save_
        Changed the purpose of the _diffrn_radiation_wavelength.id data item
        from 'Encode' to 'Key'.
 
+       Reverted _atom_type_scat.versus_stol_list to a text-type to regain
```
vaitkus, line 27670: Part of the changelog message is assigned to version 3.2.0 and part to 3.3.0. These should be moved to a single version.

cif_core.dic
```diff
@@ -25109,21 +25109,142 @@ save_atom_type_scat.versus_stol_list
          '_atom_type_scat_versus_stol_list'
          '_atom_type.scat_versus_stol_list'
 
-    _definition.update            2013-04-17
+    _definition.update            2023-04-10
```
vaitkus, line 25112: Maybe we should deprecate this item in favour of the new category?

cif_core.dic
```diff
@@ -25109,21 +25109,142 @@ save_atom_type_scat.versus_stol_list
...
+    _definition.class             Loop
+    _definition.update            2023-04-10
+    _description.text
+;
+    The CATEGORY of data items used to list atomic scattering factor values for
+    use in crystallographic structure studies.
+;
+    _name.category_id             ATOM_TYPE
+    _name.object_id               ATOM_TYPE_SCAT_TABLE
+
```
vaitkus, line 25141: @jamesrhester, can we have `ATOM_TYPE` as the parent category here? Since both the parent (`ATOM_TYPE`) and the child (`ATOM_TYPE_SCAT_TABLE`) categories are looped, wouldn't it also mean that they can be represented in a single loop? Clearly, in this case it is not possible since a single atom type can be associated with multiple entries in the `ATOM_TYPE_SCAT_TABLE`.

Or is the single-loop representation automatically rejected in this case based on the fact that `ATOM_TYPE_SCAT_TABLE` has a composite key only one part of which is linked to `ATOM_TYPE`  category key?

jamesrhester: The parent-child relationship for looped categories means that they could be joined on their key data names, which must all be common. Therefore you are right, these categories can't have a formal parent-child relationship.

jamesrhester:
I like what has been done so far on this. It looks like a definite improvement, and we should certainly deprecate the old data name, I think, and refer to the new category.

cif_core.dic
vaitkus: Ok, I changed the parent category to ATOM.

cif_core.dic
```diff
@@ -25109,21 +25109,142 @@ save_atom_type_scat.versus_stol_list
...
+    _definition.scope             Category
+    _definition.class             Loop
+    _definition.update            2023-04-10
+    _description.text
+;
+    The CATEGORY of data items used to list atomic scattering factor values for
+    use in crystallographic structure studies.
+;
+    _name.category_id             ATOM
+    _name.object_id               ATOM_TYPE_SCAT_TABLE
```
vaitkus, line 25140: Since the parent category is now atom, we could probably rename the category to `ATOM_SCAT_TABLE`.  However, I wonder if the name is not too generic. That is, do any other scattering tables come in mind instead of the stol one?

cif_core.dic
vaitkus: Ok, I deprecated the item in favour of those from the `ATOM_TYPE_SCAT_TABLE` category based on @jamesrhester's comment (https://github.com/COMCIFS/cif_core/pull/371#issuecomment-1600302774).

cif_core.dic
jamesrhester: I think it is no less generic than `atom_type_scat_table`. I'm comfortable that the new name is OK.

vaitkus: I was simply wondering maybe something like `ATOM_SCAT_VERSUS_STOL` might be better, e.g.:

```
loop_
_atom_scat_versus_stol.atom_type
_atom_scat_versus_stol.stol_value
_atom_scat_versus_stol.scat_value
Ac  0.00  89.00000
Ac  0.01  88.91457
Ac  0.02  88.66288
# ...
```

vaitkus: @jamesrhester , are you OK with the `ATOM_SCAT_VERSUS_STOL` suggestion?

jamesrhester: yes, I'm OK with that.

vaitkus: Ok, renamed the category and related data names.

vaitkus (approved):

PR #381, opened 2023-05-07
_atom_type.display_colour typo fix

rowlesmr:
Changed `colour_rgb` to `colour_RGB`.

Everything else uses uppercase RGB.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -23885,7 +23885,7 @@ save_atom_type.display_colour
     _enumeration.def_index_id     '_atom_type.symbol'
 
     _import.get                   [
-                                   {'file':templ_enum.cif  'save':colour_rgb}
+                                   {'file':templ_enum.cif  'save':colour_RGB}
                                    {'file':templ_enum.cif  'save':colour_hue}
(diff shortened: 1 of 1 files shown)


jamesrhester:
This is perhaps unexpected, but our[ dictionary style guide](https://github.com/COMCIFS/comcifs.github.io/blob/master/accepted/ddlm_dictionary_style_guide.md) points 4.3.1 and 5.1 explicitly specify all lowercase for save frame headings, and therefore by extension references to save frame headings.

Interestingly, all checks passed indicating that these particular rules are not checked by the style checker. I should fix that.

If you think this style could be improved, please raise an issue in the `comcifs.github.io` repository.

jamesrhester:
And yes, I realise that our imported save frames don't yet conform to the style rules. I'll add it to the list of issues.

rowlesmr:
so (for example) `save_pd_pref_orient_March_Dollase.diffractogram_block_id` should be `save_pd_pref_orient_march_dollase.diffractogram_block_id`?

brb changing the pd dic...

PR #387, opened 2023-05-11
lowercased all the save frame names

rowlesmr:
From https://github.com/COMCIFS/cif_core/pull/381#issuecomment-1543249332: saveframe names are supposed to be lower case.


jamesrhester:
Thank you.

Issue #391, opened 2023-05-13
templ_enum.cif: add `_units.code` to save frames


vaitkus:
It would be useful to have the units of measure in the same save frame as the enumerated values since the same save frame may potentially be imported by multiple definitions. This is probably most useful in the `atomic_mass`, `radius_bond` and `length_neutron` save frames (and potentially in `electron_count` to make it clear that is unitless and not measured in electrons).

The `_units.code` can still be retained in the importing save frames and simply ignored or replaced depending on the import mode.

jamesrhester:
I don't understand the suggestion. Perhaps you could give an example?

vaitkus:
@jamesrhester sorry for being unclear.

I suggest adding the `_units.code` attribute directly to some of the save frames in the `templ_enum.cif` file. An example of this can be seen in PR #396 where femtometres are explicitly added as units in the `length_neutron` save frame.

jamesrhester:
Yes, I see now. I agree that this is a good idea.

rowlesmr:
Questions re units.

Shouldn't all the `b` coefficients for Cromer-Mann have units of `angstrom_squared`?

![image](https://github.com/COMCIFS/cif_core/assets/14921460/99884578-ede9-4e75-bd52-5d593080f2f8)



Shouldn't the Fox coefficients all have various powers of angstroms as their units?

![image](https://github.com/COMCIFS/cif_core/assets/14921460/49ac6793-2a64-41f5-996c-7669271712cd)

Issue #317, opened 2023-01-06
[RELEASE ISSUE]: Checklist for next release of cif_core


jamesrhester:
This is a first attempt at a release checklist. We do have some things unresolved, but we'll try and deal with those as they arise. I can be release manager this time as we iron out the problems, hopefully we can rotate around.

- [x] Release manager appointed
- [x] `_dictionary_audit.revision` comments summarised
- [x] Release branch created
- [x] Next version placeholder created in dictionary_audit section in master branch
- [x] Plain text summary of changes created (ideally automatic Github changeset comments)
- [x] cif_core mailing list advised
- [x] ccp4 mailing list advised
- [x] All comments/issues arising during review period addressed (add to list below)
- [x] Final DOI and URL edited into file (IUCr Chester office to advise of DOI)
- [x] Github release tagged
- [x] Zenodo/IUCr updated with released file

Outstanding pull requests/ issues that should be resolved before releasing. Only the very easy ones go here, e.g. 'janitorial'. Please check any issues you have raised and see if they could be resolved easily or have already been resolved.

- [x] #312 
- [x] #358 
- [x] #103 
- [x] #361

jamesrhester:
Release manager is @jamesrhester this time.

vaitkus:
One question regarding the summarisation of `_dictionary_audit.revision` comments. Given that the last properly tagged version of `CIF_CORE` was 3.0.13, should all other versions (3.0.14, 3.1.0 and 3.2.0) be combined into a single new version 3.2.0 before release?

jamesrhester:
What I planned to do is summarise the comments within each `_dictionary_audit.revision` for those three versions, without actually merging them into a single version. The plain text summary would include changes from all of those three versions.

My thinking is that the dictionary audit information doesn't have to have details of all bug fixes, minor typo corrections and so on, these can just be described as "bug and typo fixes". The nature of each change will be available in the git history. More important to keep are things like new categories being added.

jamesrhester:
I have created an initial set of release notes in the wiki [here](https://github.com/COMCIFS/cif_core/wiki/Release-Notes-for-CIF-CORE-3.2.0). Please look over them and add anything you think is missing, bearing in mind that we don't want them to be so long that we lose the attention of busy software authors. @rowlesmr would you mind writing up the section on elemental composition? You could include a link to the example contained in the file.

rowlesmr:
I've added some words and links.

jamesrhester:
[...]

PR #396, opened 2023-05-15
Add units to templ_enum.cif

vaitkus:
This PR serves as an example for issue #391.


rowlesmr:
Questions re units.

Shouldn't all the `b` coefficients for Cromer-Mann have units of `angstrom_squared`?

image



Shouldn't the Fox coefficients all have various powers of angstroms as their units?

image

vaitkus:
One thing that we need to decide on before merging is if we want to only have they units in the `templ_enum.cif ` file or also in the `cif_core.dic` dictionary file. The two-place approach may be more readable, but it also violates the single point of truth principle. In other words, at some point we may update the units in the dictionary file, but forget to update them in the `templ_enum.cif`. This mismatch is quite easy to miss, since the contents are imported with the replace-on-duplicate mode which silently prefers the data items from the imported source.

@jamesrhester , @rowlesmr do you have any opinion on the topic?

jamesrhester:
I strongly believe the units should only be in one place.

A [very recent decision](https://github.com/COMCIFS/cif_core/issues/317) is that the released dictionary as it appears on the IUCr website will have all of the mode=Contents imports resolved in place, so the readability should not suffer. I am planning that the release files that Github lists will also have these resolved, but haven't got around to figuring out how that works in Github yet.

I think that means that moving the units to the `enum` file is a good idea, as it reduces the length of the main file.

jamesrhester (approved):

Issue #397, opened 2023-05-17
Impossible to index into a default list when there is more than one key value


jamesrhester:
As per discussion in #393 , to look up the scattering length of an atom both atomic mass and atomic number are required. The current `_enumeration.def_index_id` and `_enumeration_default.index` system allow only a single value to be used and so neutron scattering lengths cannot be referenced. Multiple data names could be used to specify the values for indexing by looping some new DDLm attribute, e.g. `_enumeration_index.id`, but this does not solve the issue of there being only a single column available to index into, with values given in `_enumeration_default.index`.

rowlesmr:
Just an idea of what something that works could be: (this is pretty bad, but does (could) work).

```
save_atom_type_scat.length_neutron

    _definition.id                '_atom_type_scat.length_neutron'
    _definition.update            2023-05-17
    _description.text
;
    The bound coherent scattering length for the atom type at the
    isotopic composition used for the diffraction experiment.
    
    # This definition needs to change
;
    _name.category_id             atom_type_scat
    _name.object_id               length_neutron
    _type.purpose                 Number
    _type.source                  Assigned
    _type.container               Single
    _type.contents                Real
    _enumeration.def_index_ids     
        ['_atom_type.atomic_number' '_atom_type.mass_number']
    _units.code                   femtometres

    loop_
         _enumeration_default.index_1
         _enumeration_default.index_2
         _enumeration_default.value
     1   0   -3.7390
     1   1   -3.7406
     1   2    6.671
     1   3    4.792
     2   0    3.26
     2   3    5.74
     2   4    3.26
     3   0   -1.90
     3   6    2.0
     3   7   -2.22
     #...
 
save_
```

vaitkus:
The problem here is that there is a fixed number of `_enumeration_default.index_{1,2,3,...}` items and a variable number of elements in the `_enumeration.def_index_ids` array. However, we could define the `_enumeration_default.index` item as having the same data structure as the `_enumeration.def_index_ids` array, i.e.:
```
    _enumeration.def_index_ids     
        ['_atom_type.atomic_number' '_atom_type.mass_number']
     ...
    loop_
         _enumeration_default.index_1
         _enumeration_default.index_2
         _enumeration_default.value
     [ 1   0 ]   -3.7390
     [ 1   1 ]   -3.7406
     [ 1   2 ]    6.671
     [ 1   3 ]    4.792
     [ 2   0 ]    3.26
     ...
```
[...]

PR #399, opened 2023-05-18
 index into a default list with more than one key value  (ie create `ENUMERATION_DEFAULTS`)

rowlesmr:
From #397 

This is a first attempt at putting things into place.

The idea is to allow indexing a single default value based on an arbitrary number of keys. In the example below, neutron scattering lengths are keyed on element symbol and atomic mass number. In any case, a single value is given to each combination of keys.

The definitions need work.

.

I've put in `_enumeration.def_index_ids` with a default value of `[ _enumeration.def_index_id ]`, to give backward compatibility with the current use; in the new scheme, you need to parse `_enumeration.def_index_ids`, and we don't need to redefine every single current default enumeration.

I've generalised `_enumeration_default.index` as `_enumeration_default.indexes`, even though may not make total sense to make it a plural. Again, in the new scheme, `_enumeration_default.indexes` has the default value of `[ _enumeration_default.index ]`, to give backward compatibility with the current use


rowlesmr:
An example use would be:
```
save_atom_type_scat.length_neutron

    _definition.id                '_atom_type_scat.length_neutron'
    _definition.update            2023-05-17
    _description.text
;
    The bound coherent scattering length for a given atom type and isotope.
;
    _name.category_id             atom_type_scat
    _name.object_id               length_neutron
    _type.purpose                 Number
    _type.source                  Assigned
    _type.contents                Real
    _enumeration.def_index_ids     
        ['_atom_type.element_symbol' '_atom_type.mass_number']
    _units.code                   femtometres

    loop_
         _enumeration_default.indexes
         _enumeration_default.value
     [ H   . ]   -3.7390
     [ H   1 ]   -3.7406
     [ H   2 ]    6.671
     [ H   3 ]    4.792
     [ D   . ]    6.671
     [ He  . ]    3.26
     [ He  3 ]    5.74
     [ He  4 ]    3.26
     [ Li  . ]   -1.90
     [ Li  6 ]    2.0
     [ Li  7 ]   -2.22
     #...
 
save_
```

vaitkus:
I am thinking that it might be better to create a completely new category that would be most similar to `ENUMERATION_DEFAULT`, but that would use the new multivalue-key approach (e.g. `ENUMERATION_DEFAULT_LOOKUP`, but please feel free to offer a better name). The `ENUMERATION_DEFAULT` would be deprecated and we could slowly migrate away from it in the dictionaries. The use of dREL code in the reference dictionary introduces some complexity that IMHO would be better avoided.

@rowlesmr , @jamesrhester  what are your opinions on this?

rowlesmr:
Thank makes some sense. 

Something like:
   - `_enumeration.def_index_ids` - This is an array of datanames, that, together, form the key.
   - `_enumeration_default_lookup.indexes` - This is an array of values that form the key. The values are associated to the datanames given in `_enumeration.def_index_ids`.
   - `_enumeration_default_lookup.value` - This is the single value associated with a particular combination of key values.

rowlesmr:
Just some other things to think about:

The atom symbol is a concatenation of element symbol and charge. If redesigning the entire system (which we're not), this could be changed to either separate element and charge or to also include the isotope number (if not natural abundance). 

i) Separating the charge from the element would make the atomic mass, atomic number, radius bond*, dispersion tables, and the like much simpler. This would necessitate the requirement to index multiple keys to reference scattering factors and the like.

ii) Adding the isotope number to the `_atom_type.symbol` would negate the (current) need to index multiple keys. It would also vastly expand the aforementioned tables to include all the duplicate information for each isotope. You would need an isotope number for each charge state, as structures could be shared between X-ray and neutron experiments, necessitating both scattering factors and lengths.

and then (the position which should be chosen), iii) Dont't touch `_atom_type.symbol`, and just add `_atom_type.mass_number`, and a way to index on multiple keys.


\* why are the bond radii of ions and neutral atoms the same? Whence did they come?

jamesrhester:
My thinking is as follows: If we create a new category to supercede `enumeration_default`, then we maintain perfect backwards compatibility, at the expense of carrying around an extra, unnecessary category and needing to update our `templ_enum.cif` file. If we instead change the category key of `enumeration_default` to the new list-valued object, we are not maintaining compatibility, which, with judicious use of dREL, has the following consequences:

1. Software authors who have encoded these DDLm semantics need to update software to read our updated dictionaries
2. We update version of DDLm to 5.0

Either of these approaches is viable, which means the decision is a policy one. I'm inclined to go with the update of DDLm to version 5.0, and simply provide the new category key with the dREL that explains how it is derived from the old category key. My reasons for preferring this is that at this level, very few software authors are affected (probably all are reading this message) and it keeps the foundational language cleaner.

rowlesmr:
I have no intuition as to the best approach, so I'll go with the nuclear option of just changing the existing category. If proposing that to the dlm-group, then I don't think I need to do anything more.*

I've also already updated `templ_enum.cif` to the new syntax (not on this branch, though).

\* I can't really update the scattering length, as that also needs a new data item, and that would complicate things in this PR.

vaitkus:
> Either of these approaches is viable, which means the decision is a policy one. I'm inclined to go with the update of DDLm to version 5.0, and simply provide the new category key with the dREL that explains how it is derived from the old category key. My reasons for preferring this is that at this level, very few software authors are affected (probably all are reading this message) and it keeps the foundational language cleaner.

@jamesrhester As you mentioned, there seems to be two conflicting policies here. From my point of view, the proposed approach does not seem cleaner since it breaks things for DDLm software that is not capable of dynamically interpreting dREL code. Of course, the coders can still use dREL code as a reference, but it does require going back and tweaking the software. Given that, software from the `cod-tools` package does indeed not yet handle the `ENUMERATION_DEFAULT` fields when validating CIF instance files, however, dictionary validation will also be affected by this since the software would now need to know how to dynamically construct the category key from the values of another item (as can be seen by currently issued errors during PR validation). I do acknowledge, though, that such problems probably do not arise in software that properly interprets dREL.

I also understand that having two categories which serve a similar purpose may be somewhat  confusing though still a much lesser evil from a programmers point of view as the handling of the new category can be implemented completely independently.

If the ddlm-group will prefer to go with the "nuclear" solution, I apologize in advance if the `ddlm_validate` program will break for a bit in a minor way. (It will probably be sufficient to modify the post commit check message filtering until I properly fix it).

vaitkus:
> If the `templ_enum.cif` file is updated at the same time, would that help to fix the issues `cod-tools` software might experience? The new category key would be present, the only part that would not be validated is the relationship with the new `_enumeration.def_index_ids`.

I think that would work. I would eventually implement the checking of the new relationships.

> Also, if `cod-tools` were to check the dictionary DDLm conformance version number, it could simply refuse to check anything that is not DDLm version 4 until updated.

This is indeed something worth implementing since it would at least help to detect the incompatibilities. However, the underlying need to modify the code would still remain.

> I am not wedded to the "nuclear" option, but the other option would mean an old, unused category that would ideally eventually be removed, leading again to a version 5 of DDLm.

I think that the accumulation of some outdated or deprecated definitions is unavoidable. These should preferably still be supported for a few intermediate versions and then eventually removed with the next major release (or even the next after that). Further more, is I correctly understand the current proposal, it still leaves some deprecated items (`_enumeration.def_index_id`, `_enumeration_default.index`) in the dictionary, with the only real difference being that they belong to the same category as the new items.

However, I do understand that this version of DDLm can still potentially be considered early in development so some less conservative changes may be preferred. Are these changes something you would like to see published in the upcoming volume G of ITC or are they aimed at a future release? 

As a side note, the new propose name `_enumeration_default.indexes` should indeed be changed to something else since there is still a single index, it is simply potentially a multi-column (composite) one. Maybe something like "_enumeration_default.composite_index" be better? (Though still a bit of a misnomer in cases when we have only a single column.) We could go full database and call it a `_enumeration_default.unique_index` to clearly denote a one-to-one match.

rowlesmr:
> I think that the accumulation of some outdated or deprecated definitions is unavoidable. These should preferably still be supported for a few intermediate versions and then eventually removed with the next major release (or even the next after that). Further more, is I correctly understand the current proposal, it still leaves some deprecated items (`_enumeration.def_index_id`, `_enumeration_default.index`) in the dictionary, with the only real difference being that they belong to the same category as the new items.

@vaitkus makes a good argument. I say go with this. It still ends in a v5 with things removed, but is a more gradual, and officially deprecates them prior to removal.

Side note: is there anything in the CIF release notes about communicating the actual removal of deprecated dataitems?


>As a side note, the new propose name `_enumeration_default.indexes` should indeed be changed to something else since there is still a single index, it is simply potentially a multi-column (composite) one. Maybe something like "_enumeration_default.composite_index" be better? (Though still a bit of a misnomer in cases when we have only a single column.) We could go full database and call it a `_enumeration_default.unique_index` to clearly denote a one-to-one match.

I am definitely not wedded to `_enumeration_default.indexes`. I can redo as `_enumeration_default.unique_index`.



\<FUN>
image
\</FUN>

jamesrhester:
OK, I'm happy to go with the preference of people who are actually using this thing. Let's have a new category and leave things at DDLm version 4 for now. As this is in time for Volume G 2nd Edition, I will write it up in place of the old data names. What do we think about `enumeration_defaults` as the new category name, and then `_enumeration_defaults.index` would work fine?

ddl.dic
```diff
@@ -1022,6 +1050,75 @@ save_enumeration_default.value
...
+    _name.object_id               index
+    _type.purpose                 Encode
+    _type.source                  Assigned
+    _type.container               Array
+    _type.dimension               '[]'
+    _type.contents                Code
+    _method.purpose               Definition
+    _method.expression
+;
+    _enumeration_defaults.index = [ _enumeration_default.index ]
```
jamesrhester, line 1092: This dREL can't work. It states a relationship that must be true for every row, except there is no way to indicate which row of `enumeration_default` (a different category with lots of rows) you want. The `dataname[<keyvalue?]` syntax could be used, but they you don't know the value of the key yet because that's what you're calculating.

ddl.dic
```diff
@@ -1022,6 +1050,75 @@ save_enumeration_default.value
...
+    _name.category_id             enumeration_defaults
+    _name.object_id               value
+    _type.purpose                 Encode
+    _type.source                  Assigned
+    _type.container               Implied
+    _type.contents                Implied
+    _method.purpose               Definition
+    _method.expression
+;
+    _enumeration_defaults.value = _enumeration_default.value
```
jamesrhester, line 1117: This dREL might almost work if it read:
```
_enumeration_defaults.value = _enumeration_default.value[_enumeration_defaults.index[0]]
```
except for the fact that `index` could be a mulit-entry list, in which case who knows if the first entry is the one you want to use. So I'd drop this as well.

jamesrhester:
Apart from removing the dREL I think I'm happy to merge this. What do you think @vaitkus ?

rowlesmr:
I shall remove the drel

jamesrhester:
This change can be independent of any changes in `templ_enum.cif`, as we haven't removed the old attributes.

vaitkus:
@jamesrhester There is actually one annoying issue with `_enumeration_defaults.index` which I am not sure how to resolve. The problem is that `_enumeration_defaults.index` treats all of its values as having the save content type (`Code`) while in reality the data item referenced by `_enumeration.def_index_ids` can have a whole range of content type (i.e. the `mass_number` will be an integer).

Possible solutions:
1. We could explicitly state that the content type(s) should be inherited from the items referenced by the `_enumeration.def_index_ids` attribute, but we still need some kind of content type value to indicate that this should be done. Neither `Implied` nor `ByReference` content type seem to be suitable for this task, but I am not sure a new content type should be added for just this one case.

2.a Alternatively, we could continue having a single content type `_enumeration_defaults.index`, but then it should be clearly stated that all values of the composite key are treated as if they would be of that type (e.g. even if `mass_number` is an integer, key values `01` and `1` would be treated as different). However, the question then still remains which one of the content types should be chosen? Maybe `Word`?

2.b Same as 2a, but we also state that the values are normalised before converting them to the `Word` value. However, the problem here is that we do not have any rules for normalising data values of different content types.

jamesrhester:
Yes, this type of indexing is something we have deliberately removed from the main dictionary, which wanted to index all categories using a single data name potentially created from several data values. We can't do that here without redesigning the entire default system.

We want a low-effort solution, as all we are talking about is looking up default values to inform calculations of derived values by dREL. This is not something that affects actual data transmission or archiving. So some thoughts:

1. The values used to index come from the data file, so we can't control their presentation. That means we must be able to handle caseless data values and alternative ways of writing integers.
2. We can control what datanames are used to perform the lookup (`def_index_ids`)

So I would propose simply defining an `Inherited` data type as follows:

"Used only with compound data values in the DDLm attribute dictionary. The data type of each value in a compound value is the same as that of the data name to which it is directly related."

rowlesmr:
Would you like me to work that into this PR?

jamesrhester:
Let's wait and see what @vaitkus thinks.

vaitkus:
I am generally OK with having the `Inherited` content type as long as it is only used in the reference dictionary. 

However, the relationship between the value and the related item might not always be clear to an outside reader, so the definition of `Inherited` type should strongly encourage to include rules on establishing such relationships into the human-readable definition of the data item. Of course, we will have no formal way of checking if the description actually includes such information, but it is still better than nothing. Let's update the PR along these lines and then see if it need any further fine-tuning.

In the next major revision we may come up with better ideas, but, as you mentioned, this would require a serious reworking of the typing system.

rowlesmr:
I've added some words.

The definition of `_type.container` for the value `Implied` refers to a "DDLm Reference Dictionary". Is that different to a "DDLm Attribute Dictionary"? or even a "DDLm Dictionary"?

vaitkus:
> I've added some words.

Thank you, I will have a look later on in the day, or tomorrow in the morning at the latest.

> The definition of `_type.container` for the value `Implied` refers to "the DDLm Reference Dictionary". Is that different to a "DDLm Attribute Dictionary"? or even "DDLm Dictionary"? also should it be "the" or "a" dictionary? is this supposed to refer to the file `ddl.dic`? I'm not up to speed on this level of nuance in the terminology.

To my understanding:
* The `DDLm Reference dictionary`. The single dictionary which defines the attributes used by itself and all the other dictionaries written in accordance to the DDLm language. In this repository it corresponds to the `ddl.dic` file.
* The `DDLm Attribute Dictionary`. I assume that @jamesrhester meant it as a synonym of `DDLm Reference dictionary`. I would recommend exclusively using the term `Reference` since it matches the `_dictionary.class` enumeration value. 
* A `DDLm dictionary`. Any dictionary written in accordance to the DDLm language.

vaitkus (commented):
I tried to rewrite some parts in a way that's clearer to me, but that definitely does not mean that it will be clearer for others. The specific text of the definitions is therefore still open to discussion.

ddl.dic
```diff
@@ -993,6 +993,34 @@ save_enumeration.def_index_id
...
+    _type.purpose                 Identify
+    _type.source                  Assigned
+    _type.container               Array
+    _type.dimension               '[]'
+    _type.contents                Tag
+    _method.purpose               Definition
+    _method.expression
+;
+    _enumeration.def_index_ids = [ _enumeration.def_index_id ]
+;
```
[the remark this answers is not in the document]
rowlesmr: I agree. I think is was a hang over from me trying to get away with not having to update the enumeration lists.

ddl.dic
```diff
@@ -1995,7 +2082,17 @@ save_type.contents
...
+         Inherited
+;
+         >>> Applied ONLY in a DDLm Attribute Dictionary <<<
+         For use only with List, Array, Matrix, or Table data values in the DDLm
+         attribute dictionary. The data type of each value in such a compound
+         value is the same as that of the data name to which each value is
+         directly related.
 
+         The description text of the Inherited type must include rules on
+         establishing the units of any indiviudal value in the compound data.
```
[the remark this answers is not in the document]
rowlesmr: Yes. I also had the multi-container units thing in my head, and managed to smoosh the ideas together.

ddl.dic
```diff
@@ -993,6 +993,34 @@ save_enumeration.def_index_id
...
+    _definition.id                '_enumeration.def_index_ids'
+    _definition.class             Attribute
+    _definition.update            2023-06-01
+    _description.text
+;
+    Specifies the data names of the items that, in combination, have a value
+    used as an index to the DEFAULT enumeration list (in category
+    enumeration_default) in order to select the default enumeration value for
+    that specific combination of defined items. The value of the identified data
+    items must match one of the _enumeration_default.indexes values.
```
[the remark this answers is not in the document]
rowlesmr: Roger that. How about something like:

> The value of the identified data items should be reasonably expected to match one of the _enumeration_default.indexes values.

Basically, I want to say that you can't just put random jibberish in there, it needs to be (at least) consistent jibberish.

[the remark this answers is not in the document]
jamesrhester: Well, if there is not a match, that just means that there is no default value, so I don't think we need to even mention whether or not anything has to match.

ddl.dic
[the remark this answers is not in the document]
vaitkus: Ok, so this should either be removed or changed to a definition method that assigns _enumeration.def_index_id as the default value if such attribute is present. I would probably go for the removal, but lets see if @jamesrhester has any opinion on this.

ddl.dic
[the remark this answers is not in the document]
vaitkus: Ok, so are you ok with removing this sentence (at least in this PR)?

ddl.dic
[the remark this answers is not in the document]
vaitkus: @rowlesmr Do you mean something along the lines:
```
Therefore, values of the _enumeration_defaults.index attribute inherit the
enumeration range, enumeration set and the content type of the corresponding
data item.
```

ddl.dic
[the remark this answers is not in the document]
rowlesmr: If it's like any of my other attempts at dREL, just remove it.  :)

ddl.dic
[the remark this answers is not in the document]
rowlesmr: most definitely. I think the last commit did...

ddl.dic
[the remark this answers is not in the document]
rowlesmr: I think so.  How about:

> The constituent values of _enumeration_defaults.index inherit the
  enumeration range, enumeration set, and the content type of the 
  corresponding data item.

ddl.dic
```diff
@@ -1154,6 +1182,65 @@ save_enumeration_default.value
...
+save_enumeration_defaults.index
+
+    _definition.id                '_enumeration_defaults.index'
+    _definition.class             Attribute
+    _definition.update            2023-06-01
+    _description.text
+;
+    Index key in the list default values referenced to by the value of
+    _enumeration.def_index_ids. The values given must be in the same order as
+    specified by _enumeration.def_index_ids.
```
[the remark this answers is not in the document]
rowlesmr: I tweak the above a little and added some formatting.

rowlesmr:
From my comment in a commit:

>How can a `List` dataitem have a `_type.contents` attribute, if not `Inherited`?

A `List` is:

> Ordered set of values. Elements need not be of same contents type.

I can't remember, is `_type.contents` a compulsory component of a data item?

ddl.dic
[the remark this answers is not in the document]
vaitkus: No worries, we are still all learning to deal with the beast that is dREL.

ddl.dic
```diff
@@ -1154,6 +1176,73 @@ save_enumeration_default.value
...
+
+save_enumeration_defaults.index
+
+    _definition.id                '_enumeration_defaults.index'
+    _definition.class             Attribute
+    _definition.update            2023-06-10
+    _description.text
+;
+    An index value, in a form of an ordered list, which allows the
+    identification the most suitable default value. The order of values in the
```
vaitkus, line 1204:
```suggestion
    An index value, in a form of an ordered list, which allows to
    identificy the most suitable default value. The order of values in the
```

vaitkus: Or alternatively "which allows the identification **of**".

ddl.dic
```diff
@@ -1154,6 +1176,73 @@ save_enumeration_default.value
...
+    list must correspond to the order of the related data item references in the
+    _enumeration.def_index_ids attribute list. That is, the n^th^ list element
+    is assumed to represent a value that may be assigned to the data item
+    referenced at the n^th^ position of the _enumeration.def_index_ids attribute
+    list value.
+
+    The constituent values of the _enumeration_defaults.index attribute inherit
+    the enumeration range, enumeration set, and the content type of the
+    data item from which they are paired.
```
vaitkus, line 1213:
```suggestion
    the enumeration range, enumeration set, and the content type from the
    data item with which they are paired.
```

vaitkus: The "from which they are paired" part sounded a bit weird.

ddl.dic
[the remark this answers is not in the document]
vaitkus: Thanks!

ddl.dic
```diff
@@ -993,6 +993,28 @@ save_enumeration.def_index_id
...
+    Specifies the data names of items that are collectively used to identify
+    the suitable default value in the ENUMERATION_DEFAULT category loop.
+    The values of these items are used to construct an ordered list which is
+    checked against the values of the _enumeration_defaults.index attribute.
+;
+    _name.category_id             enumeration
+    _name.object_id               def_index_ids
+    _type.purpose                 Identify
+    _type.source                  Assigned
+    _type.container               Array
```
vaitkus, line 1012:
```suggestion
    _type.container               List
```

vaitkus: I guess having an `Array` there is not incorrect, but I guess it was originally targeted more towards numeric types, mathematical operations, etc.

ddl.dic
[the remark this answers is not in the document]
vaitkus: Nope, the sentence is still there.

ddl.dic
```diff
@@ -1995,7 +2084,20 @@ save_type.contents
...
+         Intended to be used with List, Array, Matrix or Table containers which
+         may simultaneously contain values of several content types. The content
+         type of each value in such containers matches the content type of the
+         data item to which it is directly related. Specific rules for relating
+         data values to data items MUST be provided as human-readable text in
+         the _description.text attribute of all data items that have the
+         Inherited content type.
 
+         The description text of the Inherited type must include rules on
+         establishing the units of any indiviudal value in the compound data.
```
vaitkus, line 2099:
```suggestion
```

vaitkus:
> From my comment in a commit:
> 
> > How can a `List` dataitem have a `_type.contents` attribute, if not `Inherited`?
> 
> A `List` is:
> 
> > Ordered set of values. Elements need not be of same contents type.
> 
The "elements need not be of the same content type" is most likely a leftover from when the type system was quite more complex and allowed having elements of different types in the same List. However, note that this statement only says that you can define the List as having multiple types, not that any `List` is always allowed to have elements of different types. For example, if we declare the contents type of a `List` as `Integer`, that means that all list elements must be integers.

> I can't remember, is `_type.contents` a compulsory component of a data item?

The `_type.contents` has a default value of `Text` thus you can say that it is always defined. However, an explicit declaration of the `_type.contents`  attribute is highly encouraged (marked as mandatory in the `DICTIONARY_VALID` loop).

ddl.dic
rowlesmr: Your suggestion is good.

ddl.dic
rowlesmr: I'm just going from the contents definitions.

vaitkus:
> mean `[1 2 3] * [2 4 6] = [1*2 2*4 3*6]`?

Not sure if you can multiply two arrays like this. My interpretation would be more like `[1 2 3] * 5 = [5 10 15]`.

ddl.dic
vaitkus: This was resolved by removing the dREL altogether.

rowlesmr:
Sure you can. It's element-wise multiplication (Hadamad product).

jamesrhester:
Just to confirm, yes, `List`s may have different content types, but only `Implied` and `Inherited` would seem to allow that to happen. There's no point defining such a heterogeneous compound data name for use in a data file, as each of the components being of a different type means that they are processed differently which means they have a different meaning, so need a separate data name.

An `Array` has uniform content type.

A `Matrix` follows the rules of matrix multiplication, and would include items meant to be used as row/column vectors.

ddl.dic
jamesrhester: Agree with `List`.

jamesrhester:
I think this is ready for merging if those few suggested changes above are incorporated, and the minor version is bumped in `ddl.dic`, and  `_dictionary.ddl_conformance`  in `cif_core.dic` changed to match.

ddl.dic
vaitkus: Ok. Changed the container type to `List` and updated the category name in the description.

vaitkus:
@jamesrhester The last version number change already increased the minor version, that is, the version was changed from 4.0.2 to 4.1.0, so the DDL version number increase is probably not necessary. Or should we create a new version of 4.2.0 and put all the changes from this branch there?

Also, should we formally deprecate the old ENUMERATION_DEFAULT category?

jamesrhester:
Yes, we should deprecate the old category. OK not to change the version number then, as long as it is different to the one used in the 3.2.0 core release.

vaitkus:
Ok, so I did quite a few changes since the last review:
* Deprecated the `ENUMERATION_DEFAULT` category and the `_enumeration.def_index_id` attribute.
* Adapted the `ENUMERATION_SOURCE` category mechanism to work with `ENUMERATION_DEFAULTS` instead of `ENUMERATION_DEFAULT`.
* Updated the version number of DDL reference dictionary. However, it seems that we did assign some of the changes to version that were already in the 3.2.0-rc branch. I corrected the DDL reference dictionary here, but a more proper fix was implemented in PR #429. I think we should merge that PR first, sync the `master` branch with this branch and after a final review we should be good to go (IMHO).

jamesrhester (approved):

ddl.dic
```diff
@@ -994,9 +994,14 @@ save_enumeration.def_index_id
 
     _definition.id                '_enumeration.def_index_id'
     _definition.class             Attribute
-    _definition.update            2021-08-18
+    _definition_replaced.id       1
+    _definition_replaced.by       '_enumeration.def_index_ids'
+    _definition.update            2023-06-23
     _description.text
 ;
+    Deprecated. The _enumeration.def_index_id data item should be
```
rowlesmr, line 1002:
```suggestion
    Deprecated. The _enumeration.def_index_ids data item should be
```

rowlesmr (commented):
I think it looks all good. Just that one change on line 1002.

vaitkus (approved):
The validator does not yet properly handle the 'Inherited' type, but I updated the GitHub actions to ignore the warnings for now so we can start using these dictionary attributes in other feature branches.

I will most likely implement the proper handling in the upcoming few weeks.

PR #406, opened 2023-06-02
Add `ENUMERATION_SOURCE`

### LINKED ISSUE #402: Add ability to formally record enumeration source?
From https://github.com/COMCIFS/cif_core/pull/400#issuecomment-1560348338, I don't think there is the ability to firmally record the source of an enumeration value.

How about the following for addition to `ddl.dic`:
   - `_enumeration.source` for when there is a single source for an enumeration, or there is only a single value
   - `_enumeration_default.source` for when there are different sources for each value in an enumeration.
?

These tags would go in, for example, `templ_enum.cif` to record whence the values came.

```
# not a loop dataitem
save_enumeration.source

    _definition.id                '_enumeration.source'
    _definition.update            2023-05-24
    _description.text
;
    Reference to source of value(s) used in this enumeration.
;
    _name.category_id             enumeration
    _name.object_id               source
    _type.purpose                 Describe
    _type.source                  Recorded
    _type.contents                Text
    _description_example.case     'International Tables Vol. C Table 4.4.4.1'

save_
```

```
# loop dataitem
save_enumeration_default.source

    _definition.id                '_enumeration_default.source'
    _definition.update            2023-05-24
    _description.text
;
    Reference to source of value used in this enumeration for this key.
;
    _name.category_id             enumeration_default
    _name.object_id               source
    _type.purpose                 Describe
    _type.source                  Recorded
    _type.contents                Text
    _description_example.case     'International Tables Vol. C Table 4.4.4.1'

save_
```

rowlesmr:
Will close #402 

Add ability to record the source of the values in a default enumeration (ie the ones in `templ_enum.cif`) to be recorded.


ddl.dic
```diff
@@ -1002,6 +1002,26 @@ save_enumeration_default.index
...
 
+save_enumeration_default.source_id
+
+    _definition.id                '_enumeration_default.source_id'
+    _definition.update            2023-06-02
+    _description.text
+;
+    A code to identify the source material relevant to the
+    _enumeration_default.value. This code must match a corresponding
+    _enumeration_source.id.
```
vaitkus, line 1013:
```suggestion
    An identifier which corresponds to an entry in the ENUMERATION_SOURCE
    category loop. Used to provide a reference to the original source of
    a specific enumeration value.
```

vaitkus (commented):
Thank you for the work. I suggested a few rewrites which seem to bring the definitions closer to other definitions in the dictionary. Everything else seems ok, but I will take a more proper look in the evening.

ddl.dic
```diff
@@ -1076,6 +1096,62 @@ save_enumeration_set.state
...
+    _name.category_id             enumeration_source
+    _name.object_id               id
+    _type.purpose                 Key
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Word
+
+save_
+
+save_enumeration_source.reference
```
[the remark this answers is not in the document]
rowlesmr: Added in [8a9277d](https://github.com/COMCIFS/cif_core/pull/406/commits/8a9277dc810127205171592f2f8606b24c71fe3d)

I put in my interpretation of how a single reference for an entire enumeration is supposed to work, as well as a multi-reference loop.

ddl.dic
```diff
@@ -1076,6 +1096,109 @@ save_enumeration_set.state
...
+    _type.purpose                 Key
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Word
+    _method.purpose               Definition
+    _method.expression
+;
+    _enumeration.default = Unique_id(enumeration_source.id)
+;
+
```
vaitkus, line 1180: Why do we need a dREL method here?

ddl.dic
```diff
@@ -1076,6 +1096,109 @@ save_enumeration_set.state
...
+;
+;
+         There are two sources for the default values given in the enumeration.
+         Entries marked with 'a' come from Table 4.4.4.1 in the 2nd edition of
+         Volume C of the International Tables of Crystallography. Entries marked
+         with 'b' come from Table 4.3.2.1 in the 1st edition of Volume H of the
+         International Tables of Crystallography.
+;
+;
+         _enumeration_source.reference  'doi:10.1107/S2053273322010944 Table S4'
```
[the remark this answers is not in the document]
rowlesmr: will change. needs to be on the next line.

ddl.dic
rowlesmr: so that in the event of a single source, you don't need to give an id value; it just gets magicked into existence.

vaitkus: The magicking in the case of a single value loop should already happen automatically without the use of dREL code. More specifically, there is an informal agreement that software should be able to imply the value under certain condition, one of which is the one you described (single loop packet with an arbitrary key value).

The use of `Unique_id()` function is reserved for cases when we have to deal with legacy categories for which the id item is newly assigned and thus not present in multiple existing data files (e.g. the `PUBL_AUTHOR` category) category. The use of this function is not suitable here since it also assigns arbitrary values in cases when there is more than one missing key value. This behaviour is not something that we want to allow in newly defined categories since it may lead to certain data anomalies.

rowlesmr: Roger that.

ddl.dic
```diff
@@ -1002,6 +1002,26 @@ save_enumeration_default.index
...
+    _definition.id                '_enumeration_default.source_id'
+    _definition.update            2023-06-02
+    _description.text
+;
+    An identifier which corresponds to an entry in the ENUMERATION_SOURCE
+    category loop. Used to provide a reference to the original source of
+    a specific enumeration value.
+;
+    _name.category_id             enumeration_default
+    _name.object_id               source_id
```
jamesrhester, line 1016: Aren't we now using `ENUMERATION_DEFAULTS`? (`S` on the end)

jamesrhester:
Looks fine, just needs to be rewritten in terms of `enumeration_defaults`

ddl.dic
rowlesmr: That's in #404, and not merged yet.

rowlesmr: Sorry, #399

rowlesmr:
> Looks fine, just needs to be rewritten in terms of `enumeration_defaults`

Roger that. Just need `enumeration_defaults` merged in first.

vaitkus:
@jamesrhester @rowlesmr Maybe we could merge this PR as is and then simply update it once the `ENUMERATION_DEFAULTS` changes are resolved/merged?

After all, the `ENUMERATION_SOURCE` can be used to describe the sources of both the `ENUMERATION_DEFAULT` and `ENUMERATION_DEFAULTS`.

jamesrhester:
Agreed and merged.

Issue #386, opened 2023-05-11
Auxiliary dictionary files (`templ_enum.cif` and `templ_attr.cif`) are not formatted correctly


jamesrhester:
The above files are not formatted according to the style rules. They are not currently automatically checked.

PR #407, opened 2023-06-02
Format `templ_attr.cif`

rowlesmr:
From #386

I've applied the same (hopefully) formatting requirements for `cif_core.dic` to `templ_attr.cif`, wrt which columns different things should be in.

I've also reformatted the descriptions to go up to the 80 character mark. Not completely sure if this is a good idea or not.

### DIFF
--- a/templ_attr.cif
+++ b/templ_attr.cif
@@ -26,23 +26,21 @@ save_atom_site_label
     _definition.update            2021-10-25
     _description.text
 ;
-     This label is a unique identifier for a particular site in the
-     asymmetric unit of the crystal unit cell. It is made up of
-     components, _atom_site.label_component_0 to *_6, which may be
-     specified as separate data items. Component 0 usually matches one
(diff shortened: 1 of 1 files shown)


rowlesmr:
I'm leaving `templ_enum.cif` until after all the indexing stuff, default enumeration values, and reference info is sorted.

vaitkus:
Is is possible that you made these changes in your `add_beta_ADPs` (https://github.com/rowlesmr/cif_core/tree/add_beta_ADPs)  branch instead of `main`? It seems that the PR also includes some non-formating changes (e.g. addition of the `save_aniso_betaij`). We should not merge it in this form.

On a related note, I am not sure it is worth making these changes by hand. @jamesrhester may correct me on this, but I think he already has software which brings the dictionaries up-to-code automatically. I think we should go with the automated approach if possible since it is less error-prone.

rowlesmr:
Yeah. I don't know what I did there. I was switching between branches a lot yesterday.

PR #408, opened 2023-06-02
Add `DICTIONARY_AUTHOR`

### LINKED ISSUE #377: Create a dictionary_author category
The dictionaries represent quite a large scholarly effort, and the authors/guilty parties should be acknowledged. I suggest that a dictionary_author category would provide a loop over names and orcid ids, and appear after the dictionary_audit information.

Alternatively, the authors of a particular data name definition could be acknowledged, but I think that is too granular and is in any case provided by the Git history.

rowlesmr:
Will close #377.

Added the ability to record authors of a dictionary. Does there also need to be a `DICTIONARY_AUTHOR_ROLES` category? I think that may overcomplicate things, but it's an option.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -10,7 +10,7 @@ data_DDL_DIC
     _dictionary.title             DDL_DIC
     _dictionary.class             Reference
     _dictionary.version           4.1.0
-    _dictionary.date              2023-01-13
+    _dictionary.date              2023-06-02
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/ddl.dic
     _dictionary.ddl_conformance   4.1.0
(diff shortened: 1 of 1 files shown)


ddl.dic
```diff
@@ -715,6 +715,125 @@ save_dictionary_audit.version
...
+;
+    The CATEGORY of data items used for dictionary author(s) details.
+;
+    _name.category_id             DICTIONARY
+    _name.object_id               DICTIONARY_AUTHOR
+    _category_key.name            '_dictionary_author.id'
+
+save_
+
+save_dictionary_author.address
```
vaitkus, line 734: I might be wrong, but I do not predict that the `_dictionary_author.address` will be used much. What is the predicted use case -- sent me mail by post with questions regarding the dictionary and I will get back to you in 2-4 weeks? :smile:

I would rather add the option to add the author's email address (see _publ_author.email).

ddl.dic
```diff
@@ -715,6 +715,125 @@ save_dictionary_audit.version
...
+    _name.object_id               id
+    _type.purpose                 Key
+    _type.source                  Related
+    _type.container               Single
+    _type.contents                Word
+    _method.purpose               Definition
+    _method.expression
+;
+    _enumeration.default = Unique_id(dictionary_author.id)
+;
```
vaitkus, line 779: This dREL is not needed since the id item belongs to a newly added category.

ddl.dic
rowlesmr: Ah. Yeah, copied from main, not the audit_author extension branch...

ddl.dic
rowlesmr: So because its a new category, an ID is automatically generated if it isn't there?

vaitkus: Not exactly. See the following comment (https://github.com/COMCIFS/cif_core/pull/406#discussion_r1215616453) on a similar topic.

In short, the key should be automatically generated for all categories if certain conditions are met (e.g. there is a single loop entry, the key has a arbitrary value and is not referenced by other items).

Note, that the current version of the `ddlm_validate` software used in the dictionary check actions does not yet properly handle such cases, but this is the drawback of the software and not the definitions.

rowlesmr: > In short, the key should be automatically generated for all categories if certain conditions are met (e.g. there is a single loop entry, the key has a arbitrary value and is not referenced by other items).

I did not know that. That is good behaviour.

vaitkus (commented):
Thank you for adding the data item for an email address. However, I would like to go back to my previous question -- do we really need data items like address, phone number and fax? These may be useful during the CIF/paper publication process, but not that much in a DDL dictionary which is processed and distributed in a quite different way.

I suggest that we add the minimal useful set of data items for now such as name, email, orcid. The category can always be extended in the future if needed, but I really strongly doubt if we will ever need the fax field.

ddl.dic
```diff
@@ -715,6 +715,199 @@ save_dictionary_audit.version
...
 
+save_DICTIONARY_AUTHOR
+
+    _definition.id                DICTIONARY_AUTHOR
+    _definition.scope             Category
+    _definition.class             Loop
+    _definition.update            2023-06-02
+    _description.text
+;
+    The CATEGORY of data items used for dictionary author(s) details.
```
vaitkus, line 726:
```suggestion
    Attributes used to record the dictionary author information.
```

vaitkus: Minor rewording to make it more similar to other category definitions in this dictionary (and the `PUBL_AUTHOR` definition).

rowlesmr:
I think I've sent 2 faxes in my life. Both to renew my radiation licence; they accept emails now.

.

Name, email, and orcid are a minimal set; I was just being conservative in what I wrote by copying what we already had. I'm not phased if the others go.

jamesrhester:
Agree that minimal information is required here: just name, email, orcid_id.

vaitkus (approved):

PR #415, opened 2023-06-05
add unit to allow `_units.code` to express and arbitrary mixture of units

### LINKED ISSUE #409: units for dataitems with mixed units?
Is there a unit code to be used for dataitems that have mixtures of units? or should it just not be included?

cf `_atom_type_scat.cromer_mann_coeffs`, where the units are a mixture of `none` and `angstrom_squared`

(also all the other scattering factor data items)

.

Could there be a `units.code` for `mixed`-type dataitems, where there are several different unit types? Not completely machine readable, but also shows that they aren't unitless, or arbitrary.

rowlesmr:
will close #409 

as mentioned in https://github.com/COMCIFS/cif_core/issues/409#issuecomment-1576037690

Where a dataitem consists of an arbitrary number of values, or a mixture of values with different units, use `.` to indicate that the user (either a person or a program) must decide which units to apply to any single value taken from the item.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -2325,8 +2325,9 @@ save_units.code
 ;
     A code which identifies the units of measurement.
 
-    'unspecified' is used only for List, Array, Matrix, or Table data types
-    where the constituent values have differing units.
+    The 'unspecified' code should only be used in cases when the units cannot
+    be properly expressed in DDLm. A typical example of such situation is a
+    List data item with constituent values that may have differing units.
 ;
     _name.category_id             units
     _name.object_id               code
--- a/ddl.dic
+++ b/ddl.dic
@@ -10,7 +10,7 @@ data_DDL_DIC
     _dictionary.title             DDL_DIC
     _dictionary.class             Reference
     _dictionary.version           4.1.0
-    _dictionary.date              2023-01-13
+    _dictionary.date              2023-06-10
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/ddl.dic
     _dictionary.ddl_conformance   4.1.0
@@ -2089,10 +2089,13 @@ save_units.code
 
     _definition.id                '_units.code'
     _definition.class             Attribute
-    _definition.update            2021-03-01
+    _definition.update            2023-06-10
     _description.text
 ;
     A code which identifies the units of measurement.
+
+    'unspecified' is used only for List, Array, Matrix, or Table data types
+    where the constituent values have differing units.
 ;
     _name.category_id             units
     _name.object_id               code
@@ -2607,7 +2610,7 @@ save_
 
        Unified the spelling of certain words.
 ;
-         4.1.0                    2023-01-13
+         4.1.0                    2023-06-10
 ;
        Added new 'Word' content type.
 
@@ -2633,4 +2636,7 @@ save_
        Clarified the definition of the "Version" state in _type.contents
        by explicitly specifying that is adheres to the formal grammar
        provided in SemVer version 2.0.0.
+
+       Added 'unspecified' as a valid units.code for compound data types where
+       the constituent values have differing units.
 ;
--- a/ddl.dic
+++ b/ddl.dic
@@ -2869,8 +2869,7 @@ save_
        by explicitly specifying that is adheres to the formal grammar
        provided in SemVer version 2.0.0.
 
-       Added 'unspecified' as a valid units.code for compound data types where
-       the constituent values have differing units.
+       Added the "unspecified" enumeration value for the _units.code attribute.
 
        Added ENUMERATION_SOURCE and _enumeration_default.source_id to allow the
        source of default enumeration values to be recorded.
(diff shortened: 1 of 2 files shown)


rowlesmr:
@jamesrhester There was an error in the layout check:

```
ERROR: LoadError: MethodError: Cannot `convert` an object of type Nothing to an object of type String
Closest candidates are:
  convert(::Type{String}, !Matched::String) at essentials.jl:210
  convert(::Type{T}, !Matched::T) where T<:AbstractString at strings/basic.jl:231
  convert(::Type{T}, !Matched::AbstractString) where T<:AbstractString at strings/basic.jl:232
  ...
Stacktrace:
  [1] setindex!(A::Vector{String}, x::Nothing, i1::Int64)
    @ Base ./array.jl:843
  [2] copyto_unaliased!(deststyle::IndexLinear, dest::Vector{String}, srcstyle::IndexCartesian, src::SubArray{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T, 1, Vector{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T}, Tuple{Vector{Int64}}, false})
    @ Base ./abstractarray.jl:976
  [3] copyto!
    @ ./abstractarray.jl:950 [inlined]
  [4] copyto_axcheck!
    @ ./abstractarray.jl:1056 [inlined]
  [5] Vector{String}(x::SubArray{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T, 1, Vector{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T}, Tuple{Vector{Int64}}, false})
    @ Base ./array.jl:540
  [6] convert
    @ ./array.jl:532 [inlined]
  [7] setindex!(h::Dict{String, Vector{String}}, v0::SubArray{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T, 1, Vector{Union{Missing, Nothing, Dict{String, T}, Vector{T}, String} where T}, Tuple{Vector{Int64}}, false}, key::String)
    @ Base ./dict.jl:[38](https://github.com/COMCIFS/cif_core/actions/runs/5174257734/jobs/9320391335?pr=415#step:9:39)2
  [8] get_enums(d::DDLm_Dictionary)
    @ CrystalInfoFramework ~/.julia/packages/CrystalInfoFramework/IgvxU/src/ddlm_dictionary_ng.jl:676
  [9] CapitalCheck(d::DDLm_Dictionary)
    @ Main ~/work/cif_core/cif_core/julia_cif_tools/capitalisation.jl:[54](https://github.com/COMCIFS/cif_core/actions/runs/5174257734/jobs/9320391335?pr=415#step:9:55)
 [10] lint_report(filename::String; ref_dic::String, import_dir::String, warn::Bool, no_text::Bool)
    @ Main ~/work/cif_core/cif_core/julia_cif_tools/linter.jl:67
 [11] top-level scope
    @ ~/work/cif_core/cif_core/julia_cif_tools/linter.jl:107
in expression starting at /home/runner/work/cif_core/cif_core/julia_cif_tools/linter.jl:104
Error: Process completed with exit code 1.
```

ddl.dic
```diff
@@ -2320,10 +2320,13 @@ save_units.code
...
     _definition.id                '_units.code'
     _definition.class             Attribute
-    _definition.update            2021-03-01
+    _definition.update            2023-06-10
     _description.text
 ;
     A code which identifies the units of measurement.
+
+    'unspecified' is used only for List, Array, Matrix, or Table data types
+    where the constituent values have differing units.
```
vaitkus, line 2329:
```suggestion
    The 'unspecified' code should only be used in cases when the units cannot
    be properly expressed in DDLm. A typical example of such situation is a
    List data item with constituent values that may have differing units.
```

vaitkus: I suggest we generalise it a bit more.

Should we maybe also use this enumeration value as a placeholder in situations when the units are determined using dREL calculations (e.g. `_refln.a_calc`)? Software that properly handles dREL would simply override this values whereas less capable software could at least properly discern that the `_units.code` attribute was not accidentally left out.

ddl.dic
```diff
@@ -2865,6 +2868,9 @@ save_
        by explicitly specifying that is adheres to the formal grammar
        provided in SemVer version 2.0.0.
 
+       Added 'unspecified' as a valid units.code for compound data types where
+       the constituent values have differing units.
```
vaitkus, line 2872: This may need to change if the suggestion to generalise the purpose of `unspecified` is accepted.

ddl.dic
rowlesmr: Can't disagree with that. (but also not having (knowingly) worked with software that does proper dREL),

ddl.dic
```diff
@@ -2865,6 +2869,9 @@ save_
...
```
vaitkus, line 2873:
```suggestion
       Added the "unspecified" enumeration value for the _units.code attribute.
```

vaitkus (approved):
Looks good to me. Let's see if @jamesrhester has any additional comments.

jamesrhester:
Looks good to me.

PR #419, opened 2023-06-15
Change the units of some data items to 'unspecified'.

### LINKED ISSUE #409: units for dataitems with mixed units?

vaitkus:
Closes #409.

Specifically, this PR changes the units of some data items to 'unspecified'. The items in need of this change were taken from the a comment by @rowlesmr in issue #409 (https://github.com/COMCIFS/cif_core/issues/409#issuecomment-1574719776).

However, I did not modify two of the items:
* `_atom_type_scat.versus_stol_list`. This item will be overhauled to return its compatibility with DDL1 (see #216).
* `_refln.form_factor_table`. Here the units seems to depend on the type of radiation. If I correctly understand the dREL code in case of "neutron" we end up with `femtometres` and otherwise we end up with `none`. So maybe we could have a `Definition` method  that assigns units based on this criteria (as is already done in the case of `_refine_diff.density_max_su`).

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2023-06-03
+    _dictionary.date              2023-06-16
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.1.0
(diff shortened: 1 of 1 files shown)


jamesrhester (approved):
Happy with this.

rowlesmr (approved):

PR #420, opened 2023-06-16
add _dictionary.doi

### LINKED ISSUE #416: Do we need a `_dictionary.DOI` attribute?
We are planning to assign a DOI to each dictionary as well as each version of each dictionary. Is there any point including this DOI within the dictionary itself, e.g. through a `_dictionary.doi` tag?

Reasons why it would be pointless is that, if you have the actual dictionary already, the DOI doesn't give you anything new. Reasons why it might be reasonable is that the DOI could point to a landing page that provided more background information, (in which case the DOI is not really the DOI of the dictionary file itself, however.)

Similar arguments apply to `_dictionary.url`, so I wonder if that is necessary as well.

rowlesmr:
Will close #416

It is "The" DOI, as there should only be one (AFAIK).

It's "the DOI for this dictionary", as it should point to the actual dictionary file (https://github.com/COMCIFS/cif_core/issues/416#issuecomment-1594442167)

It's of type.content `Word`, as the DOI probably shouldn't be given as a URI ("I would (personally) prefer that _dictionary.DOI be given as 10.1107/cifdic_000002 etc., but that does run counter to CrossRef guidance and almost-universal practice." https://github.com/COMCIFS/cif_core/issues/416#issuecomment-1593324634)

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -10,7 +10,7 @@ data_DDL_DIC
     _dictionary.title             DDL_DIC
     _dictionary.class             Reference
     _dictionary.version           4.1.0
-    _dictionary.date              2023-06-10
+    _dictionary.date              2023-06-19
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/ddl.dic
     _dictionary.ddl_conformance   4.1.0
@@ -543,6 +543,24 @@ save_dictionary.ddl_conformance
     _type.source                  Assigned
     _type.container               Single
     _type.contents                Version
+
+save_
+
+save_dictionary.doi
+
+    _definition.id                '_dictionary.doi'
+    _definition.class             Attribute
+    _definition.update            2023-06-19
+    _description.text
+;
+    The digital object identifier (DOI) for this dictionary.
+;
+    _name.category_id             dictionary
+    _name.object_id               doi
+    _type.purpose                 Identify
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Word
 
 save_
 
@@ -2842,7 +2860,7 @@ save_
 
        Unified the spelling of certain words.
 ;
-         4.1.0                    2023-06-10
+         4.1.0                    2023-06-19
 ;
        Added new 'Word' content type.
 
@@ -2876,4 +2894,7 @@ save_
 
        Added DICTIONARY_AUTHOR to allow details of dictionary authors to be
        recorded.
+
+       Added _dictionary.doi to record the persistent Digital Object Identifier
+       of a dictionary.
 ;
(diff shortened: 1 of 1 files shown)


ddl.dic
```diff
@@ -546,6 +546,24 @@ save_dictionary.ddl_conformance
...
 save_
 
+save_dictionary.doi
+
+    _definition.id                '_dictionary.doi'
+    _definition.class             Attribute
+    _definition.update            2023-06-19
+    _description.text
+;
+    The digital object identifier (DOI) for this dictionary.
```
vaitkus, line 556:
```suggestion
    The digital object identifier (DOI) of the dictionary.
```

ddl.dic
```diff
@@ -546,6 +546,24 @@ save_dictionary.ddl_conformance
 
 save_
 
+save_dictionary.doi
+
+    _definition.id                '_dictionary.doi'
```
vaitkus, line 551:
```suggestion
    _definition.id                '_dictionary.DOI'
```

vaitkus: The capitalised version of the DOI acronym seems to be used in data names throughout CIF_CORE.

ddl.dic
```diff
@@ -546,6 +546,24 @@ save_dictionary.ddl_conformance
...
+    _description.text
+;
+    The digital object identifier (DOI) for this dictionary.
+;
+    _name.category_id             dictionary
+    _name.object_id               doi
+    _type.purpose                 Identify
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Word
```
vaitkus, line 563:
```suggestion
    _type.contents                Text
```

vaitkus: We went with `Text` in all other DOI-related definitions. Probably because DOI strings can technically contain spaces ("DOI names may incorporate any printable characters, <...> which is the character set defined by Unicode) [1].

[1] https://www.doi.org/the-identifier/resources/handbook/2_numbering#2.5.1

ddl.dic
```diff
@@ -546,6 +546,24 @@ save_dictionary.ddl_conformance
...
+
```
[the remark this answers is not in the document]
rowlesmr: @nautolycus and @publcif can do battle.

ddl.dic
rowlesmr: Can see how that is wrong now.

jamesrhester:
I think `Recommended` is the default?

vaitkus:
The `_dictionary_valid.option` does have the default value of `Recommended`, but I do not think that it automatically applies to items by default, i.e. they still have to be included in the `_dictionary_valid.attributes` list.

I have created a PR for this (#428), but note that the checks currently fail due to the dictionaries not having the item.

PR #428, opened 2023-06-22
Add _dictionary.doi to the list of items recommended in dictionaries

vaitkus:
The GitHub actions currently fail due to `_dictionary.doi` data item not being included in our dictionaries. In case we do not yet have a suitable DOI to be included in the dictionaries, we can simply silence the warning by filtering it out in the https://github.com/COMCIFS/dictionary_check_action repository. Should suppress such warnings for now?


jamesrhester:
Yes, suppressing these warnings for now is a good idea. We are gradually going to be creating DOIs as we go through a formal release of each dictionary.

PR #429, opened 2023-06-23
Update dictionary version numbers and changelog messages.

vaitkus:
This is needed to clearly delineate new changes from those already in the 3.2.0-rc release branch.


jamesrhester:
Thank you for this delicate work.

Issue #425, opened 2023-06-19
Add example CIF files that illustrate the usage of the `ATOM_ANALYTICAL` category


vaitkus:
As discussed in PR #313 it would be useful to have a full example of how the `ATOM_ANALYTICAL` category could be used in a CIF file (see https://github.com/COMCIFS/cif_core/pull/313#pullrequestreview-1196636679).

rowlesmr:
I'm putting one together. I made it in v1.1 with my TOPAS CIF making stuff and manually altered it to v2.0. So it should work with no errors at all ;).

rowlesmr:
An abridged one:

PR #430, opened 2023-06-24
Create ATOM_ANALYTICAL.cif

rowlesmr:
As requested in #425

Example use of ATOM_ANALYTICAL data items.

As well as the PD_QPA_EXTERNAL_STANDARD data items.


vaitkus:
Thank you for the work!

The file is very big and contains many data items from the powder dictionary, thus it cannot be properly validated with this dictionary alone. May I suggest that the full example file be moved to the powder dictionary repository (as a PR there) and this one be trimmed only to the parts relevant for the `ATOM_ANALYTICAL` category? Something along the lines:
```
data_analytical

# ...

loop_
_atom_analytical.id
_atom_analytical.analyte
_atom_analytical.meas_id
_atom_analytical.chemical_species
_atom_analytical.chemical_species_mass_percent
 1 Fe  a   'Fe'	    49.09
 2 Si  a   'Si O2' 	10.48
 3 Al  a   'Al2 O3'	 6.02
 4 Ti  a   'Ti O2'   0.75
 5 Mn  a   'Mn'	     0.15
 6 Ca  a   'Ca O'    0.14
 7 P   a   'P'       0.454
 8 S   a   'S'       0.007
 9 Mg  a   'Mg O'    0.27
10 K   a   'K2 O'    0.014
11 Na  a   'Na'	     0.01

loop_
_atom_analytical_mass_loss.id
_atom_analytical_mass_loss.meas_id
_atom_analytical_mass_loss.percent
_atom_analytical_mass_loss.temperature
LOI1 a   9.6   698
LOI2 a  10.71  923
LOI3 a  10.99 1273

loop_
_atom_analytical_source.id
_atom_analytical_source.technique
_atom_analytical_source.equipment_make
a  XRF 'Panalytical Axios'

# ...
```
Some additional items may be needed for context, but it would be better to limit this example to a single data block with no items from the powder dictionary.  

As a side-note, I noticed that some items from the powder dictionary have values that violate the current restrictions imposed by that dictionary. I will give a few examples here of the offending items, but I think we should raise a PR with this example file in the powder dictionary repository and continue the full discussion there. Note, I do not claim that the values are bad, just that they do not fit with the current dictionary definitions which may be too limiting or incorrect:

* The `DIFFRN_RADIATION_WAVELENGTH` loop has 4 item and 3 value columns (the `_diffrn_radiation_wavelength.type` values are missing).
* The value of `_exptl_absorpt_coefficient_mu` is often given with standard uncertainties (using the parenthesis notations, e.g. 17.7460(14)), although in the dictionary this item is currently defined as a `Number` and thus cannot have standard uncertainties. However, I find it a bit strange, since according to the human-readable definition it is calculated from other measurand items and could thus potentially have an SU. Any thoughts on this?
* Data item `_pd_meas.scan_method` value 'scan' it not one of the currently known enumeration values for this item. Should it be included?
* Data item `_pd_calc.component_intensities_total` does not seems to be currently defined in the dictionary.

vaitkus (commented):
Thank you for the updated example. A comment at the start of the file would be nice (see other example file under /examples).

Also one additional question simply out of my own curiosity. The identified elements make up about 67 %, the loss on ignition at the highest measured temperature for another 11 %, so where do the remaining 78 percent go?

rowlesmr:
> The identified elements make up about 67 %, the loss on ignition at the highest measured temperature for another 11 %, so where do the remaining 78 percent go?

The missing amount is oxygen.

The Fe is actually present as Fe2O3 in the fused bead, but is reported as wt% Fe. If you convert to the equivalent oxide, you get 70.18 wt% Fe2O3, which then brings the total to 99.47 wt%. The difference from 100 is a combination of experimental and that the othere elements are also not reported as the "correct" oxide.

vaitkus:
@rowlesmr Thank you for the explanation.

PR #431, opened 2023-06-25
Add  `_exptl absorpt.coefficient_mu_su`

rowlesmr:
Prompted by discussion https://github.com/COMCIFS/cif_core/pull/430#issuecomment-1605442143

`_exptl_absorpt.coefficient_mu` was defined as `Number`, but the description of this item is given in terms of values which are `Measurand`s. `_exptl_absorpt.coefficient_mu` was changed to `Derived` `Measurand`, and the SU data item was added.

In the process, I noticed that a bunch of SU data items' `_type.source` was `Related`, which doesn't seem to link up with the description of `Related`, so I changed them all to match the `_type.source` of the data item to which the SU belongs.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -20617,7 +20617,7 @@ save_atom_site.b_equiv_geom_mean_su
          '_atom_site_B_equiv_geom_mean_su'
          '_atom_site.B_equiv_geom_mean_esd'
 
-    _definition.update            2021-03-03
+    _definition.update            2023-06-25
     _description.text
 ;
     Standard uncertainty of the equivalent isotropic atomic displacement
@@ -20628,7 +20628,7 @@ save_atom_site.b_equiv_geom_mean_su
     _name.object_id               B_equiv_geom_mean_su
     _name.linked_item_id          '_atom_site.B_equiv_geom_mean'
     _type.purpose                 SU
-    _type.source                  Related
+    _type.source                  Derived
     _type.container               Single
     _type.contents                Real
     _units.code                   angstrom_squared
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -3464,7 +3464,7 @@ save_diffrn_radiation_wavelength.value_su
          '_diffrn_radiation_wavelength_su'
          '_diffrn_radiation_wavelength.wavelength_su'
 
-    _definition.update            2023-06-25
+    _definition.update            2021-08-03
     _description.text
 ;
     Standard uncertainty of the wavelength of radiation used in diffraction
@@ -3474,7 +3474,7 @@ save_diffrn_radiation_wavelength.value_su
     _name.object_id               value_su
     _name.linked_item_id          '_diffrn_radiation_wavelength.value'
     _type.purpose                 SU
-    _type.source                  Assigned
+    _type.source                  Related
     _type.container               Single
     _type.contents                Real
     _units.code                   angstroms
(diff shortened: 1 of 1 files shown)


rowlesmr:
There is also at least one SU data item where the `_type.source` doesn't match the parent data item:  `_exptl_crystal.size_rad` is `Recorded`, but `_exptl_crystal.size_rad_su` is `Derived`, but this is hidden through the `general_su` save frame.

vaitkus:
This should be split into two separate PRs -- one for the `_exptl_adsorption.coeffiecient_mu_su` and one for the `_type.source` changes -- so these issues could be discussed and merged independently.

vaitkus:
I checked the DDL1 dictionary and  the `_exptl_absorpt_coefficient_mu` data item was also not allowed to have SU. Maybe there is a legitimate reason for that after all?

rowlesmr:
> I checked the DDL1 dictionary and the `_exptl_absorpt_coefficient_mu` data item was also not allowed to have SU. Maybe there is a legitimate reason for that after all?

The `EXPTL` category is 
>    The CATEGORY of data items used to specify the experimental work
    prior to diffraction measurements. These include crystallization
    crystal measurements and absorption-correction techniques used.

So, the value of `_exptl_absorpt.coefficient_mu` should be calculated from things you know before the diffraction experiment. This ties in with it's `Recorded` status. It still doesn't mean that it shouldn't have an SU; there are other data items in `EXPTL` that have SUs (such as `_exptl.transmission_factor_max`, which would give you a (bounding) value of mu...). 

But then, the dREL for `_exptl_crystal.density_diffrn` involves `_cell.volume`, which is a diffraction experiment thing. Although, it does also include `_exptl_crystal.density_meas` for "density measured using standard chemical and physical methods". `_exptl_crystal.density_diffrn` is the only data item (at all) to have "_diffrn" appended to it, which I'm guessing identifies it as a data item in the exptl category, which is actually made up of diffraction-related information.

Maybe related, is that `_cell.atomic_mass` doesn't have an SU, as the atomic mass is calculated from ATOM_TYPE loops, not ATOM_SITE. Which also means you can't report an atomic mass of a unit cell from a refined structure.

jamesrhester:
Because the link from `SU` to the item it is the su of is specified via `_name.linked_item_id`, the `_type.source` is `Related` instead of derived. I note that this is not actually stated anywhere in `ddl.dic`, so perhaps we should rethink that or improve the `Related` description.

Regarding `mu`, I think it is entirely reasonable to allow it to have an su.

rowlesmr:
> Because the link from `SU` to the item it is the su of is specified via `_name.linked_item_id`, the `_type.source` is `Related` instead of derived. I note that this is not actually stated anywhere in `ddl.dic`, so perhaps we should rethink that or improve the `Related` description.

This is a wide ranging change. The `general_su` save frame gives everything `_type.source   Derived`. 

I am not qualified to properly parse the description of `Related`, but it does seem to only refer to looped lists.

```
# just FYI
         Related
;
         A value or tag used in the construction of looped lists of data.
         Typically identifying an item whose unique value is the reference key
         for a loop category and/or an item which has values in common with
         those of another loop category and is considered a Link between these
         lists.
;
```

vaitkus (approved):
I think we can merge this PR as is. The `_type.source` issue can be further discussed in other issue and PRs that are already open.

PR #433, opened 2023-06-26
_type.source of many other SU items changed from `Related`

rowlesmr:
Split from #431

A bunch of SU data items' _type.source was Related, which doesn't seem to link up with the description of Related, so I changed them all to match the _type.source of the data item to which the SU belongs.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -20617,7 +20617,7 @@ save_atom_site.b_equiv_geom_mean_su
          '_atom_site_B_equiv_geom_mean_su'
          '_atom_site.B_equiv_geom_mean_esd'
 
-    _definition.update            2021-03-03
+    _definition.update            2023-06-25
     _description.text
 ;
     Standard uncertainty of the equivalent isotropic atomic displacement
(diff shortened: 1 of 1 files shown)


vaitkus:
@jamesrhester , do you have any insights on the topic? Should the SU items have the `Related` purpose or should the inherit it from the measurand item?

jamesrhester:
`_type.source` is `Related` because the su items are `_name.linked_item_id` to the dataname they are the SU of, so therefore they are `Related` to that data name. This makes sense to me, but I note this use of `Related` is not described explicitly.

rowlesmr:
This also means that every other SU item I didn't touch is wrong, as well as the `general_su` save frame in `templ_attr.cif`.

jamesrhester:
I've decided that I think `_type.source` of "same as linked data name" is appropriate (as this PR suggests). It's not a strong opinion, as in any case the information provided is redundant.

jamesrhester:
Rejecting as per discussion in #434

Issue #437, opened 2023-06-28
Are the concepts of primitive and non-primitive data still relevant in DDLm?


vaitkus:
The `_type.source` data item uses the terms "PRIMITIVE" and "NOT PRIMITIVE" to describe some of the data items. This always slightly bothered me since these terms are not properly explained nor used anywhere else in the dictionary. However, the original DDLm paper ([[1]](https://doi.org/10.1021/ci300075z)) and the dREL paper ([[2]](https://doi.org/10.1021/ci300076w)) do provide a more detailed description.

From what I understand, *primitive data items* record data that usually cannot be deduced without repeating the entire experiment such as measurements, observations (e.g. instrument settings) and decisions made in nonlinear processes, while *non-primitive* data items record derivable data that can be directly evaluated from other data.

Does DDLm still try to make a distinction between these two concepts? If no, I suggest to completely remove any references to *primitive* and *non-primitive* items.  Is yes, I suggest to update the dictionary to more properly describe the distinction between the two. My attempt at providing a better explanation will be submitted in a separate PR.

Another thing that should be taken into consideration is the use of Evalution methods with *primitive* data items. According to the dREL paper [2]:
```
Primitive data items cannot be evaluated
and therefore contain no evaluation methods.
```
Do we want to enforce and check for this restriction? A cursory look at the `CIF_CORE` dictionary revealed that this is so far respected and `Evaluation` methods are only assigned to `Derived`, `Assigned` and `Related` data items, but some violations might be masked by the import statements.

[1] https://doi.org/10.1021/ci300075z
[2] https://doi.org/10.1021/ci300076w

vaitkus:
PR  #438 with the appropriate wording changes has been recently merged.

I will check if any 'Recorded' items have associated `Evalution` methods and then close the issue if not deviations are found.

vaitkus:
I have checked the `CIF_CORE` and `CIF_PD` dictionaries and detected no such deviations.

I have also implemented this check in the `cif_ddlm_dic_check` program which is used in GitHub actions. The updated version is not yet live on GitHub, but I plan to update it soon.

Issue #434, opened 2023-06-27
What is the `_type.source` of `SU`-related data items?


rowlesmr:
From #431 and continued in #433

I started replacing `_type.source   Related` in many SU items with type sources matching that of the linked data item, as there seemed to be a mixture of different source types, and also because the `general_su` save frame gives everything `_type.source   Derived`.

Then:
@vaitkus 
> Also, I think that the `Related` purpose might have been intentionally set for the SU items to show that they are related to the measurand data items. Not a 100% sure about that, but we should investigate a bit before making that change.

@jamesrhester 
> Because the link from `SU` to the item it is the su of is specified via `_name.linked_item_id`, the `_type.source` is `Related` instead of derived. I note that this is not actually stated anywhere in `ddl.dic`, so perhaps we should rethink that or improve the `Related` description.

I am not qualified to properly parse the description of `Related`, but it does seem to only refer to looped lists.

```
# just FYI
         Related
;
         A value or tag used in the construction of looped lists of data.
         Typically identifying an item whose unique value is the reference key
         for a loop category and/or an item which has values in common with
         those of another loop category and is considered a Link between these
         lists.
;
```

jamesrhester:
I think we can decide on a rule for this here. Blanket "Derived" would seem to be good, as we either calculate SU from multiple observations, by propagation of errors, or from the square root in the case of counting statistics. In weird cases where SU is in some sense measured we can have custom values in that definition block.

rowlesmr:
> I think we can decide on a rule for this here. Blanket "Derived" would seem to be good, as we either calculate SU from multiple observations, by propagation of errors, or from the square root in the case of counting statistics. In weird cases where SU is in some sense measured we can have custom values in that definition block.

Wouldn't an SU of something `Recorded` also be `Recorded`?

jamesrhester:
I agree that the SU of a `Derived` item is `Derived`. But in what sense is the SU of a `Recorded` also `Recorded`? Even if the quantity itself is observed, the SU is only (approximately) observed if multiple measurements are taken. In the case of counting statistics, the SU is `Derived` even if the counts are `Recorded`.

With that exception, as `Derived` means "derived from the values of other data names", and the parent `Recorded` data name is by definition *not* derived from any other data names (so the SU couldn't be, e.g. if the parent is an average), we are left with `Related` in order to completely avoid committing to one or the other. Or to create a new one, `SU`, to state that this is the `SU` of another data name and dodge the issue another way.

[...]

PR #438, opened 2023-06-28
Describe primitive data items

vaitkus:
This PR rewords some of the definitions as discussed in #437.

Specifically, it properly defines the *primitive* and *non-primitive* data items, and explicitly specifies that Definition methods should not be used on primitive items.

I was not sure if `Related` items should be considered *primitive* or *non-primitive*. I guess they should be handled in the same way ass `Assigned` values (*primitive*/*non-primitive* depending on the context).

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -1814,7 +1814,7 @@ save_method.purpose
 
     _definition.id                '_method.purpose'
     _definition.class             Attribute
-    _definition.update            2006-11-16
+    _definition.update            2023-06-28
     _description.text
 ;
     The purpose and scope of the method expression.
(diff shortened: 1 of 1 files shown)


jamesrhester (approved):

jamesrhester:
I think related items if they are links between categories are just assigned, but if they are SUs then they could be anything. So the primitive/non-primitive distinction doesn't apply neatly.

vaitkus:
> I think related items if they are links between categories are just assigned, but if they are SUs then they could be anything. So the primitive/non-primitive distinction doesn't apply neatly.

Ok, we can update the `Related` description to explicitly state that. This can probably be done in the same PR in which we clarify that the `Related` state also applies to SU items (see #434).

PR #440, opened 2023-07-01
Harmonise SU descriptions.

rowlesmr:
I've removed the description of the dataitem to which the SU is linked from the description of the SU. The SU description now only references the dataitem to which it is linked.

This means that if you update the description of an item with an SU, the description of the SU is automatically correct.

eg
```
    Standard uncertainty of the mean hydrostatic pressure
    at which intensities were measured.
```
is now
```
    Standard uncertainty of _diffrn.ambient_pressure.
```

I haven't altered any definition dates, as I don't think I substansively changed anything, but I can if desired.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -125,8 +125,7 @@ save_diffrn.ambient_pressure_su
     _definition.update            2021-03-03
     _description.text
 ;
-    Standard uncertainty of the mean hydrostatic pressure
-    at which intensities were measured.
+    Standard uncertainty of _diffrn.ambient_pressure.
 ;
     _name.category_id             diffrn
     _name.object_id               ambient_pressure_su
@@ -224,8 +223,7 @@ save_diffrn.ambient_temperature_su
     _definition.update            2021-03-03
     _description.text
 ;
-    Standard uncertainty of the mean temperature
-    at which intensities were measured.
+    Standard uncertainty of _diffrn.ambient_temperature.
 ;
     _name.category_id             diffrn
     _name.object_id               ambient_temperature_su
@@ -990,8 +988,7 @@ save_cell.reciprocal_angle_alpha_su
     _definition.update            2021-03-03
     _description.text
 ;
-    Standard uncertainty of the reciprocal of the angle
-    between _cell.length_b and _cell.length_c.
+    Standard uncertainty of _cell.reciprocal_angle_alpha.
 ;
     _name.category_id             cell
     _name.object_id               reciprocal_angle_alpha_su
@@ -1047,8 +1044,7 @@ save_cell.reciprocal_angle_beta_su
     _definition.update            2021-03-03
     _description.text
 ;
-    Standard uncertainty of the reciprocal of the angle
-    between _cell.length_a and _cell.length_c.
+    Standard uncertainty of _cell.reciprocal_angle_beta.
 ;
     _name.category_id             cell
     _name.object_id               reciprocal_angle_beta_su
@@ -1104,8 +1100,7 @@ save_cell.reciprocal_angle_gamma_su
     _definition.update            2021-03-03
     _description.text
 ;
-    Standard uncertainty of the reciprocal of the angle
-    between _cell.length_a and _cell.length_b.
+    Standard uncertainty of _cell.reciprocal_angle_gamma.
 ;
     _name.category_id             cell
     _name.object_id               reciprocal_angle_gamma_su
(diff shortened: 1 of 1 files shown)


jamesrhester:
This is largely a philosophical question. When these definitions are collected into e.g. an online dictionary or Vol G, do we want them to simply refer to the data name or to describe the quantity? Or are we planning to skip explicit description of the SU data names, in which case these changes make sense? @nautolycus , do you have an opinion on this one?

vaitkus:
I second the opinion of @rowlesmr that is not really useful to have an SU definition that is virtually a copy of the measurand definition since we run (a small) risk of eventually changing one and forgetting to change the other. Having a template SU description string with a reference to the measurand item seems to not have the same drawback.

Of course, we still retain the option to have specific SU descriptions where this matters, e.g. consider the imported description of anisotropic SU items (the 'aniso_uij_su' save frame):
```
     These are the standard uncertainty values (SU) for the standard
     form of the Uij anisotropic atomic displacement components (see
     _aniso_UIJ). Because these values are TYPE measurand, the su values
     may in practice be auto generated as part of the Uij calculation.
```

rowlesmr:
> imported description of anisotropic SU items (the 'aniso_uij_su' save frame)

Yes, I haven't touched any imported save frame descriptions, which do need to be fully fleshed out to allow importation into a general data item.

vaitkus:
@nautolycus, it would indeed be nice to hear your opinion on this topic.

To summarise the earlier discussion: Is it OK to shorten the SU item descriptions so they mainly only refer to the related measurand item (e.g. "Standard uncertainty of _diffrn.ambient_pressure.") or is it better to have a more verbose definition that mimics that of the related measurand item ("Standard uncertainty of the mean hydrostatic pressure at which intensities were measured.")?

Points in favour of short descriptions with references:
* The [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) is respected. That is the core definition is located only in the measurand item, thus changing it does not require to also modify the definition of the SU item.

Points against short descriptions with references:
* There is an additional level of indirection, i.e. individual SU item descriptions are less readable on their own (e.g. when encountered in Volume G) since the definition of the measurand item needs to also be consulted.

nautolycus:
Historically I have tended to be a WET ("write everything twice"), but with the very large number of *_su data names my preference is changing. For the specific use case I'm now looking at (Volume G typeset dictionaries), even an example such as "Standard uncertainty of _atom_rho_multipole_radial_slater.zeta0." runs beyond a column width, with no obvious breakpoints. Given that the full definition contains the back link from the SU item to its parent, the way is open for an _import.get that pulls in a description text like

"Standard uncertainty of the quantity referenced by _name.linked_item_id."  [1]

or even

"Standard uncertainty of the related measured quantity."  [2]

Option [1] may seem more attractive, because the linkage is declared explicitly, but to make use of it, the user needs access to that attribute within the definition save frame. For auto-generated dictionary representations (in print or HTML), the dictionary rendering software could replace the boilerplate description by an explicit link to the parent item definition. Because there isn't a link from *.item to *.item_su, two passes are needed to locate the SU item and associate it with its parent, but I don't see that as being particularly onerous.

>  individual SU item descriptions are less readable on their own

There is, in any case, a drift in that direction (i.e. DRYness) for other reasons - the magnetic dictionary currently has e.g. 

```
save_atom_type_scat.neutron_magnetic_j0_a2

    _definition.id                '_atom_type_scat.neutron_magnetic_j0_a2'
    _description.text
;                                 See definition for _atom_type_scat.neutron_magnetic_j0_A1
;
```

jamesrhester (approved):

jamesrhester:
Suggestions [1] and [2] and the approach of this PR simply restate information that is machine-readable in the rest of the definition frame, so they are all essentially equivalent. [1] and [2] have the advantage of being able to be put away into a template file, but the PR approach is useful for readers who do not understand DDLm, which is why I prefer it.

To handle line-length woes, beyond a certain data name length, we can revert to form [1].

nautolycus:
> To handle line-length woes, beyond a certain data name length, we can revert to form [1].

I would prefer that we stick to a single approach within the dictionary, and since this PR has already carried out the necessary work (and similar formulations exist in other dictionaries) I have no objection to its being implemented. Addressing the dictionary typesetting problems will have to be managed as an application issue in any case.

vaitkus (approved):

vaitkus:
Ok, I will merge this as is since it has already undergone manual review. We can make another PR which replaces the common parts with import statements later on (as is done elsewhere). Thank you to @rowlesmr for the work!

PR #441, opened 2023-07-01
Alter CELL-related saveframes from Recorded to Derived

### LINKED ISSUE #439: `CELL`-related saveframes in `templ_attr` need updating
`CELL`-related saveframes in `templ_attr` need updating to change from `Recorded` to whatever is appropriate (`Derived`?), considering that they're now in `DIFFRN`.

rowlesmr:
will close #439 

With moving `CELL` from `EXPTL` to `DIFFRN`, the related parameters should no longer be `Recorded`.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2023-06-25
+    _dictionary.date              2023-07-01
     _dictionary.uri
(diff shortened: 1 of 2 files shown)


jamesrhester:
I don't think the move between categories of itself changes the `_type.source`, rather it is whether or not the values can be derived from other quantities that have data names defined. I suspect the original classification of `Recorded` might have been a leftover from the days when reflections were measured separately on a point detector, from which cell parameters where calculated and then used to position the gonio for each hkl. The items going into that cell parameter calculation were not necessarily captured by CIF. Anyway, happy for these to be derived, as you could argue that the contributing reflections are listed in `DIFFRN_ORIENT_REFLN`.

PR #446, opened 2023-07-11
SUs are all related

### LINKED ISSUE #434: What is the `_type.source` of `SU`-related data items?

rowlesmr:
SU data items are all of type `Related`.

Edit: remove the linking to issue #433

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -122,7 +122,7 @@ save_diffrn.ambient_pressure_su
          '_diffrn_ambient_pressure_su'
          '_diffrn.ambient_pressure_esd'
 
-    _definition.update            2021-03-03
+    _definition.update            2023-07-01
     _description.text
 ;
     Standard uncertainty of the mean hydrostatic pressure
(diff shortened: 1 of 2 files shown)


vaitkus (approved):
Looks great, thanks!

I will remove the relation to #433 before merging, since I do want to make some additional related updates to the reference dictionary before closing that issue.

PR #450, opened 2023-07-13
Add explicit definition class to items in the reference dictionary.

vaitkus:
The default class of 'Datum' is not allowed in definitions of the Reference dictionary.


jamesrhester:
Good catch!

PR #451, opened 2023-07-13
Update dREL evaluation method of the ATOM_TYPE category

vaitkus:
The values of `_atom_site.type_symbol` should be used first when determining the values of `_atom_type.symbol` since the assigned types might differ from those derived from atom labels.
Since the dREL method for  `_atom_site.type_symbol` data item also derives the values from the values of `_atom_site.label`,  the updated dREL will still yield the same result in cases when the values of `_atom_site.type_symbol` are not explicitly defined.


jamesrhester:
I don't think this method needs to change to `Definition`, but the `ATOM_SITE` method should.

PR #472, opened 2023-11-27
Rename the 'photons per second' unit to 'photons_per_second'

### LINKED ISSUE #303: "photons per second" - unit with no underscores?
I just noticed that there is a unit defined as [`'photons per second'`](https://github.com/COMCIFS/cif_core/blob/6673ac0da6f8413ce9320133d19f4b14bf7ffd0c/templ_enum.cif#L847), with spaces instead of underscores, as all other units have.

Is this intentional?

Matthew

vaitkus:
Closes #303.

The 'photons per second' units were renamed to 'photons_per_second' in the upcoming version 1.8.8 of the `cif_img.dic` dictionary [1] so the same change can be reflected in DDLm as well.

Although it is unlikely that the units will change again, we might want to wait until the release of version 1.8.8 just to be safe before merging this PR.

[1] https://github.com/yayahjb/cbf_imgcif_dictionary/tree/main

### DIFF
--- a/templ_enum.cif
+++ b/templ_enum.cif
@@ -10,7 +10,7 @@ data_COM_VAL
     _dictionary.title            COM_VAL
     _dictionary.class            Template
     _dictionary.version          1.4.9
-    _dictionary.date             2023-11-13
+    _dictionary.date             2023-11-27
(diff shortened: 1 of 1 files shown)


jamesrhester:
As the first official DDLm version of cif_img will be based on 1.8.8, I think it makes sense for this change to be in place prior to that version being generated.

PR #473, opened 2024-01-18
Add the `_atom_site_fract.symmform` and `_atom_site_aniso.symmform` data items

vaitkus:
This PR adds the `_atom_site_fract.symmform` and `_atom_site_aniso.symmform` data items (see discussion in [1]).

The definitions of these items were almost verbatim copied from the document attached in [1] with the following changes:
* Data item `_atom_site_fract.symmform` was renamed to `_atom_site.fract_symmform` (changed the placement of the dot). The category of this item was also changed from `ATOM_SITE_FRACT` to `ATOM_SITE`.
* The human-readable definition of `_atom_site_aniso.symmform` was changed to use the term "anisotropic displacement parameters" instead of "anisotropic thermal ellipsoid".
* Both items were assigned the "Assigned" `_type.source`.
* Both items were assigned the "Encode" `_type.purpose`. (I assume here that they are intended to be machine-parsable). 

Before merging this, please also note that @jamesrhester raised some reasonable concerns in [1] (e.g., the same information might be better conveyed using a CIF2/DDLm list). However, since the `_atom_site_fract.symmform` is already used by some pieces of software and already appears in the wild, I suggest that we retain the semantics of these two data items and introduce additional list items (now or later) if so desired. How does that sound?

[1] https://github.com/COMCIFS/magnetic_dic/issues/59

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CORE_DIC
     _dictionary.title             CORE_DIC
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2023-11-14
+    _dictionary.date              2024-01-18
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


jamesrhester (approved):
As noted elsewhere, these are better presented as lists but given that they are already in use we should capture them in this form as well. I will raise an issue that a list-version should be defined.

vaitkus:
Thank you for the review.

The issue that deals with the list form of these items is #484, the same will probably have to be done in the magnetic CIF dictionary as well.

PR #474, opened 2024-02-05
Update definition of the `_alias.deprecation_date` attribute.

vaitkus:
This PR attempts to specify how '.' and '?' CIF special values should be interpreted in the context of the `_alias.deprecation_date` attribute. The assignment of a default value of '.' (not deprecated) also seemed appropriate.

Suggestions are welcome.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -81,8 +81,8 @@ save_alias.deprecation_date
 ;
     Date that the aliased tag was deprecated as a definition tag.
 
-    An unquoted value of '.' or '?' indicates that the associated alias has not
-    been deprecated.
+    An unquoted value of '.' indicates that the associated alias has not been
+    deprecated.
 ;
     _name.category_id             alias
     _name.object_id               deprecation_date
--- a/ddl.dic
+++ b/ddl.dic
@@ -81,9 +81,8 @@ save_alias.deprecation_date
 ;
     Date that the aliased tag was deprecated as a definition tag.
 
-    An unquoted value of '.' indicates that the associated alias has not been
-    deprecated while an unquoted value of '?' indicates that the deprecation
-    status of an alias is unknown.
+    An unquoted value of '.' or '?' indicates that the associated alias has not
+    been deprecated.
 ;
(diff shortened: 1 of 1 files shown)


vaitkus:
The layout checker seems to fail for reasons unrelated to the changes applied to the dictionary.

jamesrhester:
I've fixed the layout checker (bug triggered by a 'nothing' value for the default).

I don't think the suggested meaning of `?` is consistent with the way CIF works. `?` should always be equivalent to the complete absence of the attribute, or in other words, the value of any missing attribute is `?`. So if `.` is the default, then `?` must also mean "not deprecated". I agree that `.` should be the default, as we have many definitions where `_alias.deprecation_date` is missing.

There is an obvious lack of any way to say 'deprecated, but we don't know when.' This should not be a problem as we have good tracking these days. For legacy issues, we'll just have to choose a date after which it was definitely deprecated.

vaitkus:
I have updated the definition to state that both `.` and `?` indicate that the value has not been deprecated, hopefully this is OK. I am happy with any interpretation of `?` as long as it is clearly specified (without a specific definition the unknown value `?` may alternatively be interpreted as "The deprecation date is not know" or "The deprecation status is not known", etc.).

The checker seems to still be failing for the same reason, probably a stale checkout of some sort?

ddl.dic
```diff
@@ -76,17 +76,21 @@ save_alias.deprecation_date
 
     _definition.id                '_alias.deprecation_date'
     _definition.class             Attribute
-    _definition.update            2013-09-08
+    _definition.update            2024-02-05
     _description.text
 ;
     Date that the aliased tag was deprecated as a definition tag.
+
+    An unquoted value of '.' or '?' indicates that the associated alias has not
```
jamesrhester, line 84: Mentioning `?` explicitly is unnecessary, I think, as the behaviour of `?` and missing attributes will always be to adopt the default value, no need to reiterate that here.

vaitkus: Ok, I removed the mention of `?`.

PR #483, opened 2024-02-20
Add dictionary URIs for deprecated data names

vaitkus:
This PR adds dictionary URI for the deprecated aliases.

The URIs were chosen to refer to the versions of the dictionary in which those specific aliases were deprecated. These URIs might need to be changed if the definition of the `_alias.dictionary_uri` attribute get reworded (see issue #481).

PR #485, opened 2024-02-23
Update definition of the _alias.dictionary_uri attribute.

### LINKED ISSUE #481: Clarify the semantics of the `_alias.dictionary_uri` attribute
The current version of the DDLm reference dictionary defines the `_alias.dictionary_uri` attribute as: 
```
Absolute URI of dictionary to which the aliased definition belongs.
```
However, the definition is a bit imprecise which prevents this attribute from being effectively used in an automated fashion. For example, the same data name may belong to several versions of the the same dictionary and sometimes with slightly different semantics (or at least different human-readable definitions). I suggest we clarify the definition along the lines of:
```
Absolute URI of the dictionary to which the aliased definition belongs.
The URI should preferably point to the latest version of the dictionary
that is known to contain the most semantically similar definition.
```
The main benefit of this would be that we could run automated checks from time to time to see if the definitions from different  dictionaries (e.g. `mmCIF` and `CIF_CORE`) still match up to the specified level (e.g. we might not require the human-readable definitions to match verbatim, but having the same enumeration ranges would be great). I specifically used "should" in the reformulated definition to indicate that the definitions may become out of sync from time to time, but that one should strive to get them in sync when possible.

Alternatively, we may anchor the URI to a different dictionary version, e.g. the version that originally defined the data item.

vaitkus:
Closes #481.

### DIFF
--- a/ddl.dic
+++ b/ddl.dic
@@ -10,7 +10,7 @@ data_DDL_DIC
     _dictionary.title             DDL_DIC
     _dictionary.class             Reference
     _dictionary.version           4.2.0
-    _dictionary.date              2024-02-05
+    _dictionary.date              2024-02-23
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/ddl.dic
     _dictionary.ddl_conformance   4.2.0
@@ -98,10 +98,11 @@ save_alias.dictionary_uri
 
     _definition.id                '_alias.dictionary_uri'
     _definition.class             Attribute
-    _definition.update            2006-11-16
+    _definition.update            2024-02-23
     _description.text
 ;
-    Absolute URI of dictionary to which the aliased definition belongs.
+    Absolute URI of the latest version of the dictionary containing the aliased
+    definition.
 ;
     _name.category_id             alias
     _name.object_id               dictionary_uri
@@ -3044,7 +3045,7 @@ save_
        by explicitly specifying that it adheres to the formal grammar
        provided in SemVer version 2.0.0.
 ;
-         4.2.0                    2024-02-05
+         4.2.0                    2024-02-23
 ;
        # Please update the date above and describe the change below until
        # ready for the next release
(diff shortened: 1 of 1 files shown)


vaitkus:
I have further updated the PR by adding the `_alias.dictionary_version` as discussed in #481. Note, that the pdbx/mmcif dictionary also has a similar attribute named  `_item_aliases.version`.

See also PR #483 for an illustration of how the `_alias.dictionary_uri` and `_alias.dictionary_version` could be used in practice.

ddl.dic
```diff
@@ -112,6 +116,25 @@ save_alias.dictionary_uri
...
 save_
 
+save_alias.dictionary_version
+
+    _definition.id                '_alias.dictionary_version'
+    _definition.class             Attribute
+    _definition.update            2024-04-04
+    _description.text
+;
+    A version identifier of the dictionary that includes the latest compatible
```
jamesrhester, line 126: This still reads to me like we are asking for the latest version of the dictionary. I prefer "The version identifier of a dictionary containing a compatible version of the aliased definition."

Otherwise as dictionary versions increase with no change in an aliased definition, we are supposed to update our alias version to maintain conformance. Better to stick a flag in at some point in that dictionary's development, and the user can navigate forward and back to see if/when the definition changes.

vaitkus: What I meant here is that the version number should point to any version which contains the latest compatible definition (e.g. if the last compatible change occurred in version 0.1.0, but there were also version 0.2.0, 0.2.1, etc., we can point to any of those). However, you are correct to say, that this would still require us to track the changes in the external dictionaries so I reworded the definition as you suggested.

However, I did write "of the dictionary" instead of "of a dictionary" since the version number is tied to a specific dictionary given in `_alias.dictionary_uri`. Feel free to change it to "a dictionary" (or suggest a change) if that is not grammatically correct.

If you are happy with the current definition, please mark the issue as resolved.

jamesrhester (approved):

Issue #488, opened 2024-04-08
Harmonisation of dictionary naming convention


nautolycus:
@vaitkus and I were musing over the wide variation in names and references to the various dictionaries, and note the following scatter of designators (G means version on GitHub (sometimes corresponds to DDL1/DDLm versions), P refers to the current latest versions on the wwPDB site (over which COMCIFS has no say). We feel it would be beneficial to harmonise these, in ways suggested below the table. I raise this as an issue in the core area, but it relates to all active dictionary projects. If we approve a harmonisation project, I suggest making changes initially to the electron density dictionary, which I'm currently revising and which has no active input from a wider community.

| dictionary | ITG1e | ITG2e | Website |  Logo | File | DDLm  _dictionary.title |
| --- | --- | --- | --- | --- | --- | ---  |
| Core | coreCIF | coreCIF |  coreCIF |  CIF_core | cif_core.dic | CORE_DIC   | 
| Powder | pdCIF | pdCIF | pdCIF | CIF_pd | cif_pd.dic |  |
|  |   |   |   |   | cif_pow.dic (G)   | CIF_POW   | 
| Modulated struct | msCIF | msCIF | msCIF | CIF_ms | cif_ms.dic | CIF_MS   | 
| Electron density | rhoCIF | rhoCIF | rhoCIF | CIF_rho | cif_rho.dic | Cif_rho   | 
| Macromolecular | mmCIF | mmCIF | mmCIF | CIF_mm | cif_mm.dic | N/A   | 
|  |   |   |   |   | mmcif_std.dic (P)   | 
| |   |   |   |    | mmcif_pdbx.dic (P)   | 
| Image | imgCIF | imgCIF | imgCIF | CIF_img | cif_img.dic | Cif_img.dic   | 
| |   |   |   |    | mmcif_img.dic (P)   | 
| Symmetry | symCIF | symCIF | symCIF | CIF_sym | cif_sym.dic | N/A   | 
| Restraints | N/A | . | . | CIF_restraints |  cif_core_restraints.dic   | 
|  |   |   |   |   | cif_restr.dic (G)   | CIF_RSTR   | 
| Twinning | N/A | twinCIF |  . | CIF_twinning  | cif_twinning.dic   | 
|  |   |   |   |   | cif_twin.dic (G)   |  CIF_TWIN   | 
| Magnetic | N/A | magCIF  |  magCIF | CIF_mag | cif_mag.dic | MAGNETIC_CIF   | 
[...]

PR #492, opened 2024-05-15
Rename Head category to 'CIF_CORE_HEAD'

vaitkus:
This PR renames the Head category as discussed in issue #488.

Note that since the Head category name is used in the import statements of multiple dictionaries, these should be updated as soon as possible after merging the PR. I will provide a list of PRs for the dictionaries in question in a separate comment.

The `ATOM_SCAT_VERSUS_STOL` data item update dates were changed since I somehow managed to set them a month in the future in my previous PR.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2024-05-02
+    _dictionary.date              2024-05-14
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


vaitkus:
PRs in other repositories that should be merged as soon as possible after merging this PR:
* https://github.com/COMCIFS/Powder_Dictionary/pull/170
* https://github.com/COMCIFS/Electron_Density_Dictionary/pull/13
* https://github.com/COMCIFS/Modulated_Structures/pull/24
* https://github.com/COMCIFS/Restraints_Dictionary/pull/4
* https://github.com/COMCIFS/Twinning_Dictionary/pull/2
* https://github.com/COMCIFS/TopoCif/pull/81
* https://github.com/COMCIFS/MultiBlock_Dictionary/pull/14
* https://github.com/COMCIFS/cif_ed/pull/8

There is also the automatically generated imgCIF dictionary (https://github.com/COMCIFS/imgCIF), but it seems to refer to a fixed CIF_CORE version.

cif_core.dic
```diff
@@ -21,19 +21,19 @@ data_CIF_CORE
...
+    _definition.id                CIF_CORE_HEAD
     _definition.scope             Category
     _definition.class             Head
-    _definition.update            2014-06-18
+    _definition.update            2024-05-15
     _description.text
 ;
-    The CIF_CORE group contains the definitions of data items that
-    are common to all domains of crystallographic studies.
+    The HEAD category encompasses groups of categories that are common to all
```
jamesrhester, line 32: I think it makes more sense to say "The CIF_CORE_HEAD category is the top-level category for all categories in the CIF_CORE dictionary."

vaitkus: Yes, I accidentally left "HEAD" from the earlier set of revisions. I will also update the overall definition as you suggested.

However, this is probably the only place that explicitly defines CORE data items as items that are `to all domains of crystallographic studies`. Should we maybe then incorporate this part into the `_description.text` of the dictionary block?

jamesrhester: I agree that there should be considerably better description in the `_description.text` of the dictionary block.

jamesrhester (approved):

PR #494, opened 2024-05-17
Update the dictionary description

vaitkus:
This PR attempts to enhance the dictionary definition with the information that is no longer present in the updated version of the HEAD save frame definition (see discussion in https://github.com/COMCIFS/cif_core/pull/492/files/abf4ea404e8344f891f276b3d25bc15819ece781#r1602812701).

Suggestion are very welcome.


jamesrhester (approved):

jamesrhester:
Initially I thought it unnecessary to explain what CIF was, but then I realised that this file could appear or be linked from outside crystallography, so it is good to have this text near the top.

Issue #499, opened 2024-08-16
Need for _refln.id items?


gmadaria:
I would like to propose _refln.id tags as _category keys_ in those categories that affect reflections. For example, using Miller indices is clearly insufficient DIFFRN_REFLN, since measurements are currently highly redundant. Even after data reduction, a structure with domains could repeat triplets of indices in the same loop. In the case of modulated structures, all the Miller indices of each reflection, in a variable number depending on the dimension of the modulation,  would be needed so the assignment of category keys would be problematic. There is a solution proposed in the new version of the ms_cif.dic dictionary that consists of using __diffrn_refln.index_m_list_ or __refln.index_m_list_ as _category keys_, so that the additional Miller indices would be a single object, a list. However, the solution is not satisfactory, since it does not eliminate the twin problem and makes life a little more complicated for developers, since the loops would have structures like::
h k l m1... one-dimensional case
1 2 3 [-1] ...
1 2 3 [2] ...

h k l m1 m2 m3... three-dimensional case
1 2 3 [4,5,6]
1 2 3 [-1,2-3]

It is true that twin_individual could be also added as a _category key_, but wouldn't it be simpler to introduce a numerical and correlative _refln.id?

jamesrhester:
I agree that `_refln.id` is needed. A basic principle of designing relational schemes is not to use natural keys (keys which have intrinsic meaning beyond simple identification), because eventually you encounter a situation where they are no longer able to serve duty as both unique identifiers and retain their natural meaning.

In the previous case of [`PUBL_AUTHOR`](https://github.com/COMCIFS/cif_core/issues/103) and related author loops, where we introduced an author identifier, software that expects author names to be unique will continue to function in those situations where author names are unique, and fail (as it would have anyway) where author names are not unique.

For, e.g. `DIFFRN_REFLN`, mmcif has already defined `_diffrn_refln.id` so creating that should be uncontroversial. mmCIF have not created the corresponding `_refln.id`, but I think the same arguments hold as for `AUDIT_AUTHOR`: software would fail anyway when hkl are repeated in the reflection loop for advanced uses such as msCIF, so there is no loss in switching the key to `_refln.id`.

gmadaria:
[...]

PR #506, opened 2024-11-12
Replace hkl with opaque identifiers for reflections

jamesrhester:
hkl are inadequate to identify reflections as twinning or incommensurate structures will cause them to be repeated. Note also that mmCIF already defines `_diffrn_refln.id`.

Changing the unique key for `refln` and `diffrn_refln` will not have any practical effect on legacy software as that software would have broken anyway in situations where hkl are repeated.

This fixes issue #499.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -3532,7 +3532,7 @@ save_DIFFRN_REFLN
 ;
     _name.category_id             DIFFRN
     _name.object_id               DIFFRN_REFLN
-    _category_key.name            '_diffrn_refln.hkl'
+    _category_key.name            '_diffrn_refln.id'
 
 save_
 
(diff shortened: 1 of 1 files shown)


vaitkus (approved):
Looks good to me!

However, I wonder why the mmCIF dictionary defines the `_diffrn_refln.id` name, but not the `_refln.id` data name.

gmadaria:
Yes, hkls are good identifiers for merged reflections (merging redundant and symmetry-equivalent reflections is necessary before any solution or structural refinement). So, for mmCIF, a complete dictionary, and assuming that all macromolecular structures are  periodic in 3D, everything is fine. Even if crystals are twinned, the twinning dictionary has implemented the solution. The problem appears for extensions of, for example, cif_core.dic. We are currently validating CIF1 and CIF2 files using the CIF2 core dictionary and appropriate extensions. It goes fine through aliasing. The problem appears during dictionary merging since the attributes for the extended tags are fixed by the first dictionary. In the case of modulated structures, we could merge in the order cif_ms.dic + cif_core.dic and all CIF files containing exclusively 3+d indices would be handled correctly. But if the file contains, as many deposited CIF files do, 3D and modulated structures, the 3D reflection list would not be valid. If the merge were cif_core.dic + cif_ms.dic, the result would be exactly the opposite. So, everything would be fine with the proper reflection identifiers. It is true that it would cause backward compatibility issues. To what extent methods, in practice, could alleviate this is something that interests us.

PR #511, opened 2025-01-16
Fix dREL.

jamesrhester:
A number of problems have accumulated in the dREL methods and functions in the current dictionary, fixed in this pull request:

- dREL for atom_site_aniso.matrix_beta referred to undefined variable 'b' instead of 'a'
- newlines cannot appear inside single-quote delimited strings
- removed newlines in unusual places
- must use uppercase when extracting atom type from atom label

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -27708,7 +27708,7 @@ save_function.seitzfromjones
           Else   {
                If (inum == 0)   m  = AtoI(c)
                If (inum == 1 and c != '/')
-               dummy = print('illegal num in symmetry xyz')
+                   dummy = print('illegal num in symmetry xyz')
                If (inum == 2)
                {
                      n  = AtoI(c)
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -663,10 +663,10 @@ save_cell.convert_uij_to_betaij
 ;
     With c  as  cell
 
-    _cell.convert_Uij_to_betaij =                              1.4142 * Pi *
-    Matrix([[ c.reciprocal_length_a, 0, 0 ],
-                [ 0, c.reciprocal_length_b, 0 ],
-                    [ 0, 0, c.reciprocal_length_c ]])
+    _cell.convert_Uij_to_betaij =
+    1.4142 * Pi * Matrix([[ c.reciprocal_length_a, 0, 0 ],
+                          [0, c.reciprocal_length_b, 0 ],
+                          [ 0, 0, c.reciprocal_length_c ]])
 ;
 
 save_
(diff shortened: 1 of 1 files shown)


cif_core.dic
```diff
@@ -663,10 +663,10 @@ save_cell.convert_uij_to_betaij
 ;
     With c  as  cell
 
-    _cell.convert_Uij_to_betaij =                              1.4142 * Pi *
-    Matrix([[ c.reciprocal_length_a, 0, 0 ],
-                [ 0, c.reciprocal_length_b, 0 ],
-                    [ 0, 0, c.reciprocal_length_c ]])
+    _cell.convert_Uij_to_betaij =
+    1.4142 * Pi * Matrix([[ c.reciprocal_length_a, 0, 0 ],
+                          [0, c.reciprocal_length_b, 0 ],
```
vaitkus, line 668:
```suggestion
                          [ 0, c.reciprocal_length_b, 0 ],
```

cif_core.dic
```diff
@@ -27613,11 +27613,13 @@ save_function.atomtype
 
        m = Len(s)
        n = 1
```
vaitkus, line 27615: It seems that the 'n' variable is no longer needed at all.

cif_core.dic
```diff
@@ -27705,11 +27707,14 @@ save_function.seitzfromjones
           Else If (c == 'z') s[axis,2] = sign
           Else   {
                If (inum == 0)   m  = AtoI(c)
-               If (inum == 1 and c != '/') dummy = print('illegal num in
-    symmetry xyz')            If (inum == 2)   {
-                                n  = AtoI(c)
-                     If(n == 5) dummy = print('illegal translation in symmetry
-    xyz')                  s[axis,3] = Mod(10.+ Float(sign*m)/Float(n), 1.)
+               If (inum == 1 and c != '/')
+               dummy = print('illegal num in symmetry xyz')
```
vaitkus, line 27711:
```suggestion
                   dummy = print('illegal num in symmetry xyz')
```

cif_core.dic
```diff
@@ -27613,11 +27613,13 @@ save_function.atomtype
 
        m = Len(s)
        n = 1
-       If (m > 1 and s[1] not in '0123456789') n = 2
-       If (m > 2 and s[2]     in '+-'        ) n = 3
-       If (m > 3 and s[3]     in '+-'        ) n = 4
+       f = ""
+       if (m > 0) f += Upper(s[0])
+       If (m > 1 and s[1] not in '0123456789') f += s[1]
```
vaitkus, line 27618: Maybe s[1] should be lowercased in case we have something like "FE+2"?

vaitkus (approved):
Looks good overall.

cif_core.dic
jamesrhester: Good idea

PR #514, opened 2025-02-17
Added terms relevant to electron diffraction.

nautolycus:
These are the contents of the cif_ed.dic dictionary signed off by the nanED standards committee for incorporation into the core.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2024-11-12
+    _dictionary.date              2025-02-17
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


jamesrhester:
Given this is quite a big chunk, and some items are specific to ED, I think we'd better run it by the core DMG.

jamesrhester:
A query about `_diffrn_detector.distance_sample`: is this assuming a flat detector and a perpendicular distance? Or a curved detector that is equidistant? As mentioned the imgCIF dictionary can cope with more complex detector configurations, just not sure what the configuration envisaged for this definition is.

lukaspalatinus:
The motivation behind this item was to provide complementary information to _diffrn_detector.ed_calibration_constant. The geometry in mind was a fixed distance of a stationary flat detector. I think that the current comment in the dictionary tried to capture this. Currently it reads: _Mean distance in millimetres from the diffracting sample to the detector. Appropriate for instrument geometries where the detector is fixed or maintained at a constant distance from the diffraction target. More general geometries and provision for adjusting the instrument geometries between or during scans are handled by the imgCIF dictionary._
Maybe we can modify the beginning to "Mean distance in millimetres from the diffracting sample to the center of the detector..." This would remove the ambiguity for tilted or curved detectors.

jamesrhester:
My concern was that this definition appears in the core dictionary so may be encountered by a wide variety of people looking for a pre-cooked definition for their experiment, so best to be very clear about what it applies to. So how about: "Mean distance in millimetres from the diffracting sample to the center of _a flat_ detector..."? 

And just double-checking: it really is supposed to be to the centre of the detector, and not the distance to the closest point on the detector? So if the sample and detector centre are misaligned, the value of this item should reflect that?

nautolycus:
I had some further discussion with Lukas and his conclusion was: 

> I have given it quite a bit of thought, trying to find a non-controversial, useful and universally acceptable solution to the problems associated with these items. Finally, I am coming to the conclusion that the best solution is to remove both these items (_diffrn_detector.distance_sample and _diffrn_detector.ed_calibration_constant) from the cif_core. It might be nice to have these values, but they really belong to the imgCIF dictionary rather than to the core. James Hester correctly pointed out the issues with the detector distance for other than the simplest geometries. And there is a similar problem with the definition of the calibration constant. Even worse, the calibration constant also has various definitions and there is not a universally established formula for it. Combining the problems with the fact that these values do not have any direct implications to the data quality or data interpretation, and make only sense in combination with the raw data, I suggest to leave them out and defer the solution of this problem to imgCIF specification.

Accordingly I have removed both items from this PR. The ED community might, as suggested, revisit them in future extensions focussing on raw data processing.

vaitkus:
@nautolycus just a minor typesetting suggestion. The DDLm version of `CIF_CORE` seems to be using proper Unicode characters in favour of CIF markup where possible (e.g. "α" instead "\a"). Maybe the `_diffrn_source.convergence_angle` data item (contains "\a") and the `_refine_ls.sample_shape_expression` (contains "\t") should also be updated accordingly?

nautolycus:
> @nautolycus just a minor typesetting suggestion. The DDLm version of CIF_CORE seems to be using proper Unicode characters in favour of CIF markup where possible (e.g. "α" instead "\a"). Maybe the _diffrn_source.convergence_angle data item (contains "\a") and the _refine_ls.sample_shape_expression (contains "\t") should also be updated accordingly?

Happy to go with that and have updated the PR accordingly.

jamesrhester (approved):

Issue #517, opened 2025-03-26
Imported site_symmetry description text not generic


publcif:
In [{'file':templ_attr.cif  'save':site_symmetry}]

_description.text refers to 'angle':

... p, q and r refer to the translations that are subsequently
    applied to the symmetry transformed coordinates to generate
    the atom used in calculating the _angle_ ...
    
Perhaps it should be 'geometric parameter' or 'distance or angle', or something similar to be applicable to e.g. _geom_bond.site_symmetry_1?

vaitkus:
Nice catch, this definitely needs to be updated. Another way of changing it would be to do away with the second part of the sentence altogether:
```
... p, q and r refer to the translations that are subsequently
applied to the symmetry transformed coordinates. These
translations (x,y,z) are related to (p,q,r) by the relations ...
```

nautolycus:
Resolved with https://github.com/COMCIFS/cif_core/pull/518

Final text reads:

```
    _description.text
;
     Data item specifying the symmetry operation codes applied to the atom
     sites involved in a specific geometric configuration or other correlated
     behaviour.

     The symmetry code of each atom site as the symmetry-equivalent position
     number 'n' and the cell translation number 'pqr'. These numbers are
     combined to form the code 'n pqr' or n_pqr.

     The character string n_pqr is composed as follows:

         n refers to the symmetry operation that is applied to the
         coordinates stored in _atom_site.fract_xyz. It must match
         a number given in _space_group_symop.id (or one of its
         aliases, such as _symmetry_equiv_pos_site_id).

         p, q and r refer to the translations that are subsequently
         applied to the symmetry transformed coordinates to generate
         the related atom position. These translations (x,y,z) are related
         to (p,q,r) by the relations
              p = 5 + x
              q = 5 + y
              r = 5 + z
;
```

Issue #516, opened 2025-03-11
_journal.data_validation_number and _journal.validation_number


nautolycus:
I am curious as to the difference between the two definitions below, and specifically why and when the second (_journal.validation_number) was added. It is absent from DDL1 and DDL2 versions of the cif core; no-one in IUCr Chester knows its purpose; it is not found in any CIFs in Chester or (according to https://project-gemmi.github.io/pdb-stats/cod-tags.html) in the COD. Since `_journal_*` have all been essentially reserved terms used in Chester, and there is no mention of the new item in the dictionary audit record, I'm very puzzled. It does seem to have been present in the earliest versions of the DDLm core on GitHub.

```
save_journal.data_validation_number

    _definition.id                '_journal.data_validation_number'
    _alias.definition_id          '_journal_data_validation_number'
    _definition.update            2021-10-27
    _description.text
;
    Journal data items are defined by the journal staff.
;
    _name.category_id             journal
    _name.object_id               'data_validation_number'
    _type.purpose                 Encode
    _type.source                  Recorded
    _type.container               Single
    _type.contents                Word

save_

save_journal.validation_number

    _definition.id                '_journal.validation_number'
    _definition.update            2013-01-23
    _description.text
;
    Data validation number assigned to journal.
;
    _name.category_id             journal
    _name.object_id               validation_number
    _type.purpose                 Encode
    _type.source                  Assigned
    _type.container               Single
    _type.contents                Code

save_

```

jamesrhester:
I have searched through the early versions of the core dictionary supplied to me by Syd Hall in 2013 and have found the following changes:

```
     1.4.05   2012-12-11
;
   Changed computing.data_collection to computing.diffrn_collection.
   Changed computing.data_reduction to computing.diffrn_reduction.
   Added the CITATION, CITATION_AUTHOR, CITATION_EDITOR definitions.
;
# -- Edited out some other entries --
     1.4.09   2013-01-23
;
   Change _journal.data_validation_number to _journal.validation_number
;
```

I deduce from this that Syd was against the use of the generic word `data` in data names, as seen from the first entry above, but he forgot to add an alias and delete the old data name in the case of `_journal.validation_number`.

I suggest that, given that this "new" tag is used nowhere, we can remove it. Or, if you prefer, remove the old one and add an alias, as has been done with the computing tags above it.

vaitkus:
The issue was resolved by merging PR #518.

PR #518, opened 2025-03-27
Made _journal.data_validation_number an alias of …

nautolycus:
_journal.validation_number in cif-core.dic (addressing https://github.com/COMCIFS/cif_core/issues/516) and made some style/formatting changes in templ_attr.cif, including adoption of the suggested wording change to  [{'file':templ_attr.cif 'save':site_symmetry}] posted by @publcif in https://github.com/COMCIFS/cif_core/issues/517

I am starting to look at typesetting the core CIF dictionary for ITG 2nd edition. This process will take some time, but will throw up deviations from wording in the First Edition or deviations from house style that I would like to correct. These should be essentially cosmetic, but I always appreciate a review from a keener pair of eyes.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -17962,24 +17962,6 @@ save_journal.coden_cambridge
 
 save_
 
-save_journal.data_validation_number
-
-    _definition.id                '_journal.data_validation_number'
-    _alias.definition_id          '_journal_data_validation_number'
-    _definition.update            2021-10-27
-    _description.text
-;
-    Journal data items are defined by the journal staff.
-;
-    _name.category_id             journal
-    _name.object_id               'data_validation_number'
-    _type.purpose                 Encode
-    _type.source                  Recorded
-    _type.container               Single
-    _type.contents                Word
-
-save_
-
 save_journal.issue
 
     _definition.id                '_journal.issue'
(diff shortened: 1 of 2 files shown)


vaitkus:
Looks good overall, but I do have a few comments:
1. In the `save_aniso_bij` the reference to external bibliography was updated to not use the '[1], [2]' notation. However, the same definition still contains one '[1]' at the end of the formula line (<...> Note that <...>).
2. There are several other definitions that use the '[1], [2]' notation to refer to external sources (e.g. `save_aniso_betaij`, `save_aniso_uij` and potentially others). These should probably be updated to use the preferred way of citing sources.
3. The `_journal.data_validation_number` and `_journal.validation_number` had a few semantic differences:
  3.1. _type.source: `Recorded` vs. `Assigned`.  I think "Assigned" is the more correct one here so no change is needed.
  3.2.  _type.contents: `Word` vs. `Code`. The only difference here is case-sensitivity. That is, `_journal.data_validation_number` became case-insensitive. Is this change intended? If not, I would probably retain the old content type, i.e. change the type in the newly defined data item.
4. The change to '_site_symmetry' looks good (and it may be refined later if needed), but it should also be kept in mind that this save frame is also imported by dictionaries other than `CIF_CORE`. For example, it is currently imported by the `_restr_U_rigid.site_symmetry_1` data item from the `CIF_RSTR` dictionary (https://github.com/COMCIFS/Restraints_Dictionary). Not sure if  this also falls under "geometric parameter" or if something more generic should be used.

vaitkus (commented):
I left a few notes as a general comment.

nautolycus:
> I left a few notes as a general comment.

Great - very helpful.

1, 2. Cleaned the remaining '[1]' style references. grep doesn't find any more in templ_attr.cif or cif_core.dic, but I'll be mindful of this when going through other dictionaries too.

3. Change to "Word" accepted. Not sure that it makes a difference in the real world, but is probably the right choice.
4. Excellent point. I didn't follow your original simpler solution of truncating the sentence because I did want to retain the idea of a generated atom (strictly atom site). I hope the suggested fix is sufficiently general without being too obscure!

vaitkus (approved):
Looks good now, thanks!

nautolycus:
Closed with https://github.com/COMCIFS/cif_core/commit/abc0787719a77339b44f2ee9f986cca9eef4f494

PR #527, opened 2025-05-16
Editorial style changes to many definitions.

nautolycus:
Another round of minor changes to style as I work through the dictionary entries for Volume G typesetting.


vaitkus (approved):
There seem to be some minor merge conflicts relating to the dictionary date and changelog messages, but other than that looks good to me.

PR #532, opened 2025-06-06
Spelling error in _chemical_formula.moiety

nautolycus:
(looks like a regexp substition of "Single" for "char" !)


vaitkus (approved):
Nice catch!

Issue #56, opened 2017-10-20
ddl.dic: proposal for the expansion of the 'Real' type definition


merkys:
I suggest expanding the definition of ``Real`` to allow fractional numbers by allowing to append ``/[1-9][0-9]*`` to currently allowed ``Real`` numbers. Such expansion would allow to avoid rounding problems such as conversion of ``1/3`` to ``0.333333``, what is somewhat a loss of information – it is not unusual for papers to report coordinates/occupancies in fractional numbers.

Of course such change would require all reading programs to be aware of possible fractional numbers. However, specialized libraries already exist for [Perl](http://search.cpan.org/~davecross/Number-Fraction-2.00/lib/Number/Fraction.pm), [Python](https://docs.python.org/2/library/fractions.html) and [Java](http://commons.apache.org/proper/commons-math/userguide/fraction.html) languages.

jamesrhester:
Unfortunately such a change would render currently deployed CIF-reading software unable to correctly interpret data values that used this new convention, so it would cause a painful transition for relatively little gain, particularly as atoms at special positions can be detected in other ways anyway.

merkys:
This convinces me. The issue can be closed.

PR #554, opened 2025-09-08
Updated reference and example in _cell.formula_units_Z and *_details

nautolycus:
No change to version or date as this essentially completes the earlier posting with final page references and a requested example.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -916,7 +916,7 @@ save_cell.formula_units_z
     in exceptional circumstances (Brock, 2025) may be reported as
     non-integral.
 
-    Reference: Brock, C. P. (2025). Acta Cryst. A81, nnn-nnn.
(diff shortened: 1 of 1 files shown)


nautolycus:
OK. I was initially reluctant to have numeric examples for Z, as the appearance of a non-integral one might encourage casual users to overlook its 'exceptional' nature; but I'm won over by your thoughts about validation. Are the additions satisfactory? I seem to recall that from time to time queries arise about defining a "rational" (or "fractional") type. This is something I don't want to pursue here, but are there any open (or closed) issues about this that you are aware of?

nautolycus:
Ah just spotted closed issue https://github.com/COMCIFS/cif_core/issues/56 - enough said (for now :)

vaitkus (approved):

PR #555, opened 2025-09-15
Remove incorrect alias for diffrn_radiation_wavelength.id

jamesrhester:
`_diffrn_radiation_wavelength.id` is not an alias of `_diffrn_radiation.wavelength_id`, instead the latter is a pointer to the former.


vaitkus:
That is quite an unfortunate naming convention (two different data items with names that only differ in the placement of the dot). However, not much we can do as this is inherited from mmCIF and imgCIF dictionaries.

vaitkus (approved):

PR #557, opened 2025-09-21
Corrected the definition of a* in _atom_site.B_iso_or_equiv and similar

nautolycus:
Following correspondence with Massimo Nespolo (18-09-2025); He also provided a link to his paper  [Nespolo, M. (2024). _J. Appl. Cryst._ **57**, 1733-1746] for more detailed explanation:

> Dear Brian,
> 
> Sorry for the late answer, semester has started here.
> The problematic sentence is indeed the one where a*, b* and c*  are called ‘reciprocal-space cell lengths’. They are the 'reciprocal of the direct-space cell lengths', which do not coincided with the  ‘reciprocal-space cell lengths' unless the unit cell is primitive. The fact that the reciprocal of a unit cell is not a unit cell (unless the latter is primitive) has gone unnoticed for so long time that I was really surprised when I "discovered" it. But now that it has been pointed out, the imprecise sentences require correction.
> 
> Best wishes.
> 
> Massimo

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2025-09-18
+    _dictionary.date              2025-09-21
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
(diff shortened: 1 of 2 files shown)


jamesrhester:
This PR is now replaced by #592 and https://github.com/COMCIFS/Attribute_Templates/pull/7/ with no changes.

PR #561, opened 2025-10-22
Changes to modulated structures imports for consistency.

jamesrhester:
These changes are necessary for the updates to the MS dictionary to be consistent with our checking. I'm currently working on them [here](https://github.com/jamesrhester/Modulated_Structures/tree/gm_update_2) and will create a PR when all the tests pass.

These changes fall into two groups:
* Some definitions have a `_definition.update` in both template block and main definition, which causes an import error. I have adopted the principle that by default the change dates are in the main dictionary, unless both `_description.text` and `_method.expression` for the definition are found in the template block. This is a simple heuristic for working out where the substantial information is. In the days of Git, I don't think this attribute matters much, as Git gives us much more fine-grained change information.

* The current DDLm checks assert that the presence of `_method.expression` implies a value is derived, so some import blocks need a `_type.source` of `Derived`. All of these cases have methods showing how to pull a value out of a matrix, or put it in, so, given that a value can be calculated from other data names (if present), `Derived` makes sense, even if the methods are circular.

### DIFF
--- a/templ_attr.cif
+++ b/templ_attr.cif
@@ -154,13 +154,12 @@ save_matrix_pdb
 
 save_matrix_w
 
-    _definition.update           2023-09-11
     _description.text
 ;
      Element of the matrix W defined by van Smaalen (1991); (1995)
 ;
     _type.purpose                Number
-    _type.source                 Assigned
+    _type.source                 Derived
     _type.container              Single
     _type.contents               Real
     _enumeration.default         0.0
@@ -169,7 +168,6 @@ save_matrix_w
 
 save_ms_index
 
-    _definition.update           2014-06-27
     _description.text
 ;
      Additional Miller indices needed to write the reciprocal vector
@@ -177,14 +175,13 @@ save_ms_index
      _diffrn_standard_refln.index_m_list, _exptl_crystal_face.index_m_list.
 ;
     _type.purpose                Number
-    _type.source                 Recorded
+    _type.source                 Derived
     _type.container              Single
     _type.contents               Integer
      save_
 
 save_q_coeff_element
 
-    _definition.update           2024-08-06
     _description.text                   
 ;
      Element of the matrix _atom_site_Fourier_wave_vector.q_coeff.
@@ -192,7 +189,7 @@ save_q_coeff_element
      However, in magCIF the number of elements is limited to 3.
 ;
     _type.purpose                Number
-    _type.source                 Assigned
+    _type.source                 Derived
     _type.container              Single
     _type.contents               Integer
     _enumeration.default         0
@@ -201,7 +198,6 @@ save_
 
 save_q_coeff_seq_id
 
-    _definition.update           2024-07-26
     _description.text                   
 ;
      Element of the matrix _atom_site_Fourier_wave_vector.q_coeff_seq_id.
@@ -209,7 +205,7 @@ save_q_coeff_seq_id
 ;
     _name.linked_item_id         '_cell_wave_vector.seq_id'
     _type.purpose                Link
-    _type.source                 Related
+    _type.source                 Derived
     _type.container              Single
     _type.contents               Integer
     _enumeration.range           0:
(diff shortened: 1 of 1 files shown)


vaitkus:
This looks good to me, but there are some alternative solutions we could adopt:
- The import logic defaults to 'Exit' on duplicate items, but we could explicitly specify to 'Replace' or 'Ignore' such items (this should be specified in the modulation dictionary). I haven't tested it in this specific case, but it should work (if it does not, I should fix it then).

On a somewhat related note, having dates in both the dictionary and the template file might make sense since it makes it easier to detect and track where the specific changes occur. Not sure which should take precedence in the "merged" file, thought, since both files may introduces important changes. Unfortunately, there is no option to choose the later of the two. 

- The `_type.source` check, while legitimate, might be a bit of overzealous in this case. The check was implemented based on the following definition of the `Evaluation` method type from [1]:
```
    Method evaluates an item from related item values. Definitions of
    primitive data items should normally not contain methods of this type.
```
In this case, both the standalone value data items and the matrix data item are 'Assigned' and thus primitive. However, the 'calculations' involved in their retrieval are simple matrix construction/value retrieval operations which are semantically different from calculating the value based on different data items. I guess, it is not a big problem if we leave either the matrix or all individual values as 'Assigned', but marking them all as 'Derived' just to satisfy the checker seems somewhat incorrect. Also, note that the definition states 'should normally not contain', maybe to accommodate situations just like these. In the future, it might make sense to somehow distinctly mark methods that operate on Primitive values, but only to retrieve them from other data structures with Primitive values, e.g. by introducing a new method type or placing a specific dREL comment in the method to inform the checker to simply ignore these cases. Of course, the checker could also try to detect such relations automatically, although, since it currently does not use a proper dREL parser, the approach could be error-prone. For now, I can think of three alternative approaches:
1. Update the GitHub action to filter out these specific error messages. This would then fix the issue both in the modulated_cif repository and any repositories that import it.
2. I could modify the underlying `cif_ddlm_dic_check` script to recognise specific dREL comment (e.g. `'# cif_ddlm_dic_check disable=NO_EVALUATION_OF_PRIMITIVES'`) and ignore such deviations as explicitly allowed by the dictionary maintainer. A very similar approach is used by the `shellcheck` and `perlcritic` utilities.
3. I could modify the underlying `cif_ddlm_dic_check` script to accept an option that explicitly allows certain primitive items to have Evaluation methods. Although this option should then be somehow included in the specific repositories that require it such as the Modulated_CIF dictionary repo.

What are your thoughts on this?

[1]  https://github.com/COMCIFS/cif_core/blob/02b65fa614a4d087a428caa0e3e96892a6c50b8b/ddl.dic#L1837

jamesrhester:
> This looks good to me, but there are some alternative solutions we could adopt:
> 
>    * The import logic defaults to 'Exit' on duplicate items, but we could explicitly specify to 'Replace' or 'Ignore' such items (this should be specified in the modulation dictionary). I haven't tested it in this specific case, but it should work (if it does not, I should fix it then).

The default `Exit` was quite useful, as that is how I picked up that `_type.source Derived` that I'd added to the main dictionary contradicted the values in the template. Because the update date has no semantic significance, removing it from a definition doesn't actually change anything, and we retain 'Replace' and 'Ignore' for those cases where we do wish to override part of a definition. So I'd be in favour of adopting the principle that the date appears in one place only. If we didn't have 'git blame' I might think differently.

Regarding `Derived`:

>  For now, I can think of three alternative approaches:
> 
>    1. Update the GitHub action to filter out these specific error messages. This would then fix the issue both in the modulated_cif repository and any repositories that import it.

I think this is the best option.

>    2. I could modify the underlying `cif_ddlm_dic_check` script to recognise specific dREL comment (e.g. `'# cif_ddlm_dic_check disable=NO_EVALUATION_OF_PRIMITIVES'`) and ignore such deviations as explicitly allowed by the dictionary maintainer. A very similar approach is used by the `shellcheck` and `perlcritic` utilities.

Embedding information like this in comments in the dictionary itself seems to me to be mixing informational levels. I'd rather the dictionary was a pristine object that was understandable given only the standards and scientific background. 

>    3. I could modify the underlying `cif_ddlm_dic_check` script to accept an option that explicitly allows certain primitive items to have Evaluation methods. Although this option should then be somehow included in the specific repositories that require it such as the Modulated_CIF dictionary repo.

That sounds like a recipe for creating a maintenance nightmare as special cases accumulate.

My general thoughts are that we must conclude that an `Evaluation` method does not necessarily imply `_type.source Derived`, even though that sounds reasonable. So I would be in favour of removing this check, or printing a warning that doesn't cause everything to fail.

Issue #247, opened 2021-09-14
Licensing dictionaries


jamesrhester:
Assuming the copyright for all the dictionaries lies with the IUCr (separate issue actually), under what terms are the dictionaries made available? Currently we have a 2000-era [IUCr policy](https://www.iucr.org/resources/cif/comcifs/policy) and no licence.

This issue is related to use of the dictionaries by the [EMMO project](https://github.com/emmo-repo/CIF-ontology/issues/61). They themselves seem happy with a CC-BY-4.0 licence.

k00ni:
Any news here?

jamesrhester:
Thanks for bumping this. Conversation now [under way](https://www.iucr.org/__data/iucr/lists/comcifs-l/msg00771.html).

merkys:
> Thanks for bumping this. Conversation now [under way](https://www.iucr.org/__data/iucr/lists/comcifs-l/msg00771.html).

It seems that a vote had taken place in COMCIFS and [CC-BY has been declared the winner](https://www.iucr.org/__data/iucr/lists/comcifs-l/msg00802.html). Can we take this result as a sign CC-BY license now applies to dictionaries within this repository since vote announcement? It would be nice to have a `LICENSE` or `COPYING` file in the repository as is usual, or at least a mention in `README.md`.

A legal nitpick: there are several versions of CC-BY license. Which one of them has been had in mind in the COMCIFS discussion?

publcif:
CC-BY-4.0

vaitkus:
I suggest that we also include an SPDX license ID [1] comment into the individual dictionary files given that they are often distributed outside of the GitHub repository (e.g. via the IUCr website) and that there is currently no formal way in the DDLm dictionary to specify the license. It would look something like:
```
#\#CIF_2.0
# SPDX-License-Identifier: CC-BY-4.0
##############################################################################
#                                                                            #
#                      DDLm REFERENCE DICTIONARY                             #
#                                                                            #
##############################################################################
```
If this is suitable, it would be nice to do this before the next release.

[1] https://spdx.dev/learn/handling-license-info/

merkys:
SPDX headers are really nice, thanks. Top-level LICENCE would also be great to have. The latter is sort of a bigger thing, as usually such file is taken as a sign that all files under this repository are covered by it (unless of course said files have their own license pointers).

vaitkus:
[...]

PR #570, opened 2025-11-25
Add a LICENSE file with the plain text of CC-BY-4.0

vaitkus:
This PR licenses all files in the repository under CC-BY-4.0 as discussed in issue #247.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,395 @@
+Attribution 4.0 International
+
+=======================================================================
+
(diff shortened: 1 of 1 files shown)


vaitkus:
@jamesrhester @nautolycus I wonder if the CIF2-EBNF.txt can be (re)licensed under CC-BY-4.0 given that it was first published as supplementary material to the CIF 2.0 publication (https://journals.iucr.org/j/issues/2016/01/00/aj5269/index.html).

PR #573, opened 2025-12-10
Update definition of _audit_support.funding_organization_doi

vaitkus:
Resolves issue #275.

The note was added based on the current version of https://www.crossref.org/documentation/member-setup/constructing-your-dois/suffixes-containing-special-characters/. The linked advice could be summarised as "Encode `#` as `%23` and do NOT encode some other specific symbols, but in practice it seems that URL-encoding any symbol still results in the URI being resolved. As such, I left the note slightly less specific since symbols like ` ` or `%` will also always need to be escaped (although they probably do not appear in DOIs as often as `#`).

The content type was changed from `Text` to `Word` do disallow space characters (and thus force them to be URL-encoded).

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2025-12-09
+    _dictionary.date              2025-12-10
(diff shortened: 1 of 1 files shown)


jamesrhester (approved):
I note the unfortunate need to include the resolver in this particular DOI data name for historical reasons (definition already published) as discussed in #275 .

PR #582, opened 2026-02-02
Add definition for AMCSD database code.

### LINKED ISSUE #569: [New data item]: _database_code.AMCSD
### Definition

Code assigned by the American Mineralogist Crystal Structure Database (AMCSD).

### Restricted values

### Example

0000001

### Explanation of example

Unique identifier assigned to a crystal structure entry in the AMCSD (https://www.rruff.net/odr/amcsd/0000001).

### Looping

top level

### Data name

_database_code.AMCSD, _database_code_AMCSD

### Type

Word (text with no spaces)

### Data structure

### Comments

Crystal structures files distributed by the AMCSD have been including the dotless alias of the data item for some time now. Since AMCSD data is open, this data name also appears in other composite datasets. Therefore, I suggest that adding it retroactively might be the best approach.

From what I have seen, all AMCSD identifiers consist of 7 digits which may include leading zeros.

jamesrhester:
Closes #569

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -17483,6 +17483,24 @@ save_DATABASE_CODE
 
 save_
 
+save_database_code.amcsd
+
+    _definition.id                '_database_code.AMCSD'
+    _definition.update            2026-02-02
+    _description.text
+;
+    Code assigned by the American Mineralogist Crystal Structure Database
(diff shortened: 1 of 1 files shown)


vaitkus (approved):
Thanks!

PR #585, opened 2026-02-16
Clarify meaning of InChi version

### LINKED ISSUE #551: cif_core.dic: _chemical.identifier_inchi_version definition may be imprecise
The definition of this item states:
```
    Version number of the InChI standard to which the associated
    chemical identifier string applies.
```
It also provides the following example: `1.03`

I think there is a slight confusion here between the _InChI identifier version_ (aka _InChI standard version_) and the _InChI software version_. If I understand correctly, currently there is only a one InChI standard release (version 1) while there are multiple InChI software releases (`1.03` is one of them, see https://en.wikipedia.org/wiki/International_Chemical_Identifier#Software).

Since InChI software is the reference implementations (and potentially, the only true implementation) in a transient sense it does define the version of the InChI standard. However, if a release only includes some optimisation and memory leak fixes, calling it a different version of the standard would not make a lot of sense. I suggest that the definition be reworded to call this the "InChI standard version" instead of the "InChI standard version". Of course, we may also raise this question on the InChI Trust mailing list or the their GitHub repository (https://github.com/IUPAC-InChI) to get a definitive answer.

As for the InChI standard version, there is no need to include it as a separate item, since the major version number ("1") is already embedded both in InChI and in InChIKey values.

jamesrhester:
Fixes #551.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2026-02-02
+    _dictionary.date              2026-02-16
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


vaitkus (approved):
Looks good! Sorry for not creating this PR myself.

PR #590, opened 2026-04-09
Add an instrument DOI identifier

### LINKED ISSUE #589: [New data item]: Add support for PIDINST DOIs
### Definition

Within the context the [Research Data Alliance (RDA)](https://www.rd-alliance.org/), the [PIDINST work group](https://www.rd-alliance.org/groups/persistent-identification-instruments-wg/activity/) have developed a metadata schema for describing an instrument.  The group maintain their assets through a [GitHub organisation](https://github.com/rdawg-pidinst/) with corresponding repos, one of which is their [documentation of their work](https://docs.pidinst.org/en/latest/).

The DataCite DOI registry has worked with the PIDINST group to build a profile for the DataCite metadata that allows for the minting of DOIs to represent instruments.

Some similar concepts already exist in CIF core.  The `_diffrn_source.beamline` attribute is likely the closest (at least, for large-scale facilities), but `_diffrn_source.make` is potentially also captured in the PIDINST metadata schema.

In general, it would be helpful if CIF core supported a standard way of indicating the DOI of the instrument that generated the data.

Note that, although this issue is focused on DOIs, PIDINST also supports identifiers that are an ePIC.  In the future, additional identifiers may be supported.

### Restricted values

### Example

10.5442/ni000019

### Explanation of example

The DOI `10.5442/ni000019` has a corresponding URL `https://doi.org/10.5442/ni000019` using the DOI resolver.  This describes the "CPMU17_EMIL" is a hard x-ray EMIL beamline at HZB's BESSY II facility.

### Looping

top level

### Data name

_diffrn_source.instrument_doi

### Type

Text (arbitrary length text)

### Data structure

### Comments

_No response_

jamesrhester:
This fixes #589.

Note the definition text attempts to address the scope of the DOI, as components of instruments may also be assigned instrument DOIs (e.g. detectors), but the intention of this dataname is to encompass the entire instrument. There's probably a precise way to express how far up the component hierarchy you go: maybe as far as possible without including any "instruments" that would introduce ambiguity as to the source of the data, so at a lab you could include an integrated X-ray source as part of the instrument but at a facility you would not include the synchrotron ring (as that includes all instruments on the ring), only the monochromator. Is that explanation worth including?

I've also left scope for future DOI schemes beyond the RDA recommendations, by not specifying that the DOIs should conform to the DataCite/RDA schema.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -5894,7 +5894,9 @@ save_diffrn_source.instrument_doi
     _definition.update            2026-04-09
     _description.text
 ;
-    A DOI for the instrument used to measure the data.
+    A DOI for the instrument used to measure the data. This DOI
+    typically refers to a facility beamline or a laboratory instrument,
+    rather than to a specific component.
 ;
     _name.category_id             diffrn_source
     _name.object_id               instrument_DOI
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -5888,6 +5888,31 @@ save_diffrn_source.facility
 
 save_
 
+save_diffrn_source.instrument_doi
+
+    _definition.id                '_diffrn_source.instrument_DOI'
+    _definition.update            2026-04-09
+    _description.text
+;
+    A DOI for the instrument used to measure the data.
+;
+    _name.category_id             diffrn_source
+    _name.object_id               instrument_DOI
+    _type.purpose                 Encode
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Text
+
+    _description_example.case     10.5442/ni000019
+    _description_example.detail
+;
+        Data were collected at the hard X-ray EMIL beamline at the BESSY
+        synchrotron. Metadata for this DOI conform to the RDA recommended
+        schema at https://doi.org/10.15497/RDA00070.
+;
+
+save_
+
 save_diffrn_source.make
 
     _definition.id                '_diffrn_source.make'
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2026-02-26
+    _dictionary.date              2026-04-09
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
@@ -29034,7 +29034,7 @@ save_
        Changed the purpose of the _diffrn_radiation_wavelength.id data item
        from 'Encode' to 'Key'.
 ;
-         3.3.0                    2026-02-26
+         3.3.0                    2026-04-09
 ;
        # Please update the date above and describe the change below until
        # ready for the next release
@@ -29205,4 +29205,6 @@ save_
        Add data item for PaNET experimental method identifier.
 
        Added the _dictionary.licensing_SPDX dictionary attribute.
+
+       Add data item for instrument DOI.
 ;


vaitkus (approved):

cif_core.dic
```diff
@@ -5888,6 +5888,33 @@ save_diffrn_source.facility
 
 save_
 
+save_diffrn_source.instrument_doi
```
paulmillar, line 5891: (Sorry, if this is my fault!)

RDA PIDINST group defined the metadata schema and mapped this to two persistent identifier (PID) frameworks: [ePIC](https://www.pidconsortium.net/) and [DataCite](https://datacite.org/).

ePIC uses Handles as persistent identifiers, while DataCite (as a DOI registry) mints DOIs.

As a concrete example, [the b2inst service](https://b2inst.gwdg.de/) provides an easy way for someone to obtain an ePIC for their instrument.

As it happens, both ePIC and DataCite/DOIs use Handles; however, I would suggest that's largely a coincident.  Someone could (at least, in principle) map the PIDINST metadata to some other, non-Handle-based PID technology.

That's just a long way of saying that it might make sense to use a more technology-neutral term.  I think this would be achievable, provided the nature of the PID (e.g., epic vs doi) is encoded.

    doi:10.5442/ni000019

or

    epic:21.11157/ffr63-ps520

(Please don't use the epic example: it's a randomly chosen, but valid PID.)

This comment also applies to other places in the document where "DOI" is mentioned, e.g.,

     _name.object_id               instrument_DOI

cif_core.dic
```diff
@@ -5888,6 +5888,33 @@ save_diffrn_source.facility
...
+
+    _definition.id                '_diffrn_source.instrument_DOI'
+    _definition.update            2026-04-09
+    _description.text
+;
+    A DOI for the instrument used to measure the data. This DOI
```
paulmillar, line 5897: This could be true, but I think (for crystallographic measurements) individual instruments, such as diffractometer. 

Here is a randomly chosen example:

https://www.stoe.com/products/stadivari/

This instrument could be registered in B2INST service.  I think would be good if the corresponding Handle could be also included in the CIF file.

cif_core.dic
```diff
@@ -5888,6 +5888,33 @@ save_diffrn_source.facility
...
+    rather than to a specific component.
+;
+    _name.category_id             diffrn_source
+    _name.object_id               instrument_DOI
+    _type.purpose                 Encode
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Text
+
+    _description_example.case     10.5442/ni000019
```
paulmillar, line 5908: If we want to support non-DOI (but PIDINST-compatible) PIDs in the same field then we may need to identify the type of the identifier; for example, including the ePIC PIDs from the B2INST service.

(As it happens, we _could_ identify DOIs from ePICs because they use the same Handle technology, but other PIDs might use a different scheme.)

As an example; we could use `doi` to identify DOI PIDs and `epic` to identify ePIC PIDs.

    doi:10.5442/ni000019
    epic:21.11157/ffr63-ps520

(please don't use the epic example: it's a randomly chosen, but valid PID.)

paulmillar (commented):
Basically, do we want to limit this to DOIs, or include other PIDs

cif_core.dic
jamesrhester: I see the point. However, including the handle type inside the data value creates internal structure that software has to extract which goes against the philosophy of capturing all distinct items of information in separate data names.

Therefore we should create two data names, e.g. `_diffrn_source.instrument_pid` and `_diffrn_source.instrument_pid_framework`.

But maybe we want to accommodate multiple frameworks, so we loop:
```
loop_
    _diffrn_source_instrument_pid.pid
    _diffrn_source_instrument_pid.framework
21.11157/ffr63-ps520     epic
1234567.abcd             doi
```
And other aspects of the setup could have pids, so probably what we really want is a loop as follows:
```
loop_
    _diffrn_pids.pid
    _diffrn_pids.framework
    _diffrn_pids.component
21.11157/ffr63-ps520 epic instrument
123456/abcd          doi   source
11.12.13.14          doi   detector
```
where component and framework are drawn from a list.

cif_core.dic
jamesrhester: See my above comment.

cif_core.dic
jamesrhester: A further question: suppose we allow a range of different PIDs as I've written above. Should we identify the metadata schema behind them, e.g. PIDINST, or is that discoverable from the PID?

cif_core.dic
paulmillar: I think the approach of having two identifiers: one for the ID itself (`_diffrn_source_instrument_pid.pid`) and the other for the nature of the PID (`_diffrn_source_instrument_pid.framework`) makes a lot of sense.

It's similar to the approach adopted by DataCite (see [RelatedIdentifier](https://datacite-metadata-schema.readthedocs.io/en/4.7/properties/relatedidentifier/)'s `relatedIdentifierType`)

On whether to focus on the overall source or whether to support subcomponents (e.g., the source, the detector) , which (I believe) underpins the `_diffrn_source_instrument_pid` vs `_diffrn_pids` point ...

Please note that PIDINST metadata has some support for identifying subcomponents: the metadata may include the identifiers of subcomponents.  Although I believe this would provide some overlap with the `component` concept, I believe it doesn't necessarily allow the reader to understand the nature of the subcomponent within the overall instrument.

If the `component` approach is preferred, then it would be helpful if the documentation made it clear which `component` value would be appropriate for a beamline (or end-station) located at a large-scale facility; e.g., P23, located at PETRA III.

cif_core.dic
paulmillar: I think there's broadly two use-case for PIDs
  1. linking related resources.
  2. discovering more information about the instrument.

The first use-case doesn't directly depend on metadata (although, perhaps when an instrument has multiple identifiers.)

For the second use-case, I don't think we have sufficient experience to know what are the use-cases we should support.  Moreover, completely describing the metadata is actually quite complicated.

My suggestion would be to leave this out for now.  We could add some metadata scheme field once we have a clearer picture of level of compatibility we're trying to achieve.  This should also help us understand how to describe the metadata schema.

jamesrhester:
This pull request superceded by #598.

PR #592, opened 2026-04-10
Fix reciprocal axis terminology

jamesrhester:
This is simply that content of #557 relating to `cif_core.dic`. A separate PR in the other repository will cover changes to `templ_attr.cif`. Once I've prepared that I will close #557.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -10,7 +10,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2025-09-18
+    _dictionary.date              2025-09-21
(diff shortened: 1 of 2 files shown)


vaitkus:
@jamesrhester Maybe this and the corresponding PR in the template dictionary should also be merged before the next release of coreCIF given that they are already prepared?

I will approve the PR, but it would be good if someone more experienced could confirm that the change is scientifically sound.

vaitkus (approved):

jamesrhester:
I have reviewed Nespolo, J. Appl. Cryst. (2024) p 1733. The change in terminology is to reflect the fact that the unit cell of the reciprocal lattice will not be described by the reciprocals of the direct space unit cell if the direct space unit cell is non-primitive.

PR #595, opened 2026-04-21
Update GitHub action dependencies

vaitkus:
GitHub is deprecating Node 20 on GitHub Actions runners [1] therefore some actions will need to be migrated to newer versions. I have also used this opportunity to update the step names and clean up the layout of the action file.

I have tested the action on my branch and it seems to work properly.

[1] https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,7 @@ jobs:
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
         - name: checkout
-          uses: actions/checkout@v4
+          uses: actions/checkout@v6
           
       # Check syntax of all CIF files
         - name: check_syntax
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -65,79 +68,81 @@ jobs:
...
-                 
+          path: ~/.julia
+          key: ${{ runner.os }}-julia-v2
+
       - name: Install Julia
-        uses: julia-actions/setup-julia@v1
+        uses: julia-actions/setup-julia@v3
         with:
-                version: '1.6'
+          version: '1.9'
```
jamesrhester: Julia version 1.10 would be better as that is LTS.

jamesrhester (approved):
Updated Julia version to 1.10 which is LTS.

vaitkus:
@jamesrhester Yes, using the 1.10 LTS makes more sense. I copied 1.9 from another template (cif_topo), so I will update it there as well within the PR that migrates away from Node 20.

As for the stale cache, I missed it since I had previously manually removed the cache. To avoid similar issues in the future, I will update the GitHub action to dynamically include the Julia version and maybe even a hashsum of the Julia tools dependency file into the cache name.

PR #598, opened 2026-04-29
Add datanames to capture instrument PIDs

### LINKED ISSUE #589: [New data item]: Add support for PIDINST DOIs

jamesrhester:
This PR replaces #590 and fixes #589. See initial discussion there. This PR provides datanames for capturing any PIDs associated with the instrumentation. I have made the following choices when preparing these three datanames:

1. The PID doubles as a category key. This assumes PIDs from different sources will not collide. Seems reasonable. The true category key would include `framework` as well, but we can add that to the key when an actual collision seems likely.
2.  I have restricted the possible frameworks using an enumerated list, which gives guidance to users and ensures conformity.
3.  A 'framework' includes both the PID type (DOI etc.) and the metadata specification. In theory those are two different things being conflated in a single value but I don't see much advantage in having those two pieces of information separate. Anybody using these datanames will want the metadata so will need both the PID type and schema to get at that.
4. The `details` dataname is provided as a simple human-readable description and is not intended for classification of instrumentation which would be a very very deep rabbit hole.

I may have messed up terminology ('register' PID, 'framework') so feel free to correct it. I'm pinging @paulmillar just in case.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -2502,7 +2502,10 @@ save_DIFFRN_COMPONENT_PID
 ;
     _name.category_id             DIFFRN
     _name.object_id               DIFFRN_COMPONENT_PID
-    _category_key.name            '_diffrn_component_pid.pid'
+    loop_
+      _category_key.name
+         '_diffrn_component_pid.pid'
+         '_diffrn_component_pid.framework'
 
     _description_example.case
 ;
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2026-04-20
+    _dictionary.date              2026-04-29
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


paulmillar:
The proposed change looks good to me.

cif_core.dic
```diff
@@ -2488,6 +2488,114 @@ save_diffrn_attenuator.scale
...
+    _type.source                  Assigned
+    _type.container               Single
+    _type.contents                Text
+
+    loop_
+      _enumeration_set.state
+      _enumeration_set.detail
+         DataCite
+;
+         DOI resolving to metadata described in the RPAINST white paper
```
vaitkus: Is the RPAINST acronym correct? Unfortunately, I am not familiar with it, and could not locate it in the cited white paper.

cif_core.dic
```diff
@@ -2488,6 +2488,114 @@ save_diffrn_attenuator.scale
...
+         at 10.15497/RDA00070 and using the DataCite schema.
+;
+         ePIC
+;
+         An ePIC handle resolving to metadata matching the RPAINST schema
+         defined at DOI 10.1547/RDA00070.
```
vaitkus: This DOI does not seem to resolve, maybe it should have been 10.15497/RDA00070?

cif_core.dic
```diff
@@ -2488,6 +2488,114 @@ save_diffrn_attenuator.scale
...
+    _definition.update            2026-04-29
+    _description.text
+;
+    The category of data items providing persistent identifiers for
+    instrumentation associated with data collection. These identifiers
+    will generally be registered with independent organisations.
+;
+    _name.category_id             DIFFRN
+    _name.object_id               DIFFRN_COMPONENT_PID
+    _category_key.name            '_diffrn_component_pid.pid'
```
vaitkus: I would really recommend making this a composite key (framework + pid) from the start. If/when an identifier collision occurs, it is already too late since multiple files might have already be created without the explicit framework, thus leading to ambiguities in data files.

cif_core.dic
paulmillar: Ah, sorry, I missed that.  This should be "PIDINST white paper": PIDINST is "Persistent Identification of Instruments". 

If its helpful, you could also call it the "RDA PIDINST white paper", where RDA is the Research Data Alliance.

cif_core.dic
paulmillar: Yes, the correct DOI is `10.15497/RDA00070`.  (Sorry, I thought I had checked the DOIs.)

cif_core.dic
paulmillar: Handle-based PIDs (ePIC, DOI, ...) and URL-based PIDs can't collide, provided the URL has a FQDN in its authority (e.g., `my-pids.example.org` in the URL `https://my-pids.example.org/foo`).  A non-Handle and non-URL based PID _could_ collide (with either a Handle or a URL) but assigning such PIDs a new framework value would ensure there is no ambiguity.

I believe the framework is not optional.  However, even if it is omitted, it seems very unlikely that any new PID framework would use identifiers that are neither URLs nor Handles, but are somehow ambiguous with these existing schemes.

The additional information (that the proposed change doesn't included at this stage) is about the "metadata specification".

To me, there's two potential causes for concern:

  1. the existing metadata standard is updated in a non-backwards compatible fashion,
  2. a PID technology (e.g., DOI) is used in two PID schemes with mutually incompatible schemes.

Both are possible but seem rather unlikely to me.

The nearest possible would be the practice of using a paper (journal article) that describes a beamline.  The paper would have a DOI, but not a DataCite DOI (conforming to PIDINST) but a CrossRef DOI (with no relation to PIDINST).  This isn't a use-case for which I would advocate and I don't know of anyone who is (or might) propose this.    (Even in this case, it is possible to determine between DataCite or CrossRef DOIs.)

In the description, the proposed solution would be to add an additional framework.

Just to illustrate the idea, I've created some examples using a `+` to combine the two items:

  - `DOI`  for DOIs conforming to PIDINST,
  - `DOI+foo`   for DOI conforming to metadata schema "foo",
  - `DOI+bar`   for DOI conforming to metadata schema "bar",
  - `ePIC` for ePICs conforming to PIDINST,
  - `ePIC+baz` for ePICS conforming to metadata schema "baz".

Theoretically, someone _could_ create CIF data with framework `DOI` and with the PID conforming to metadata schema "foo" (i.e., not PIDINST).   However, per the documentation, this is clearly wrong.

Therefore, I don't believe there is an issue with the current proposal.

I believe you are advocating that we start by explicitly identifying PIDINST, something similar to writing `DOI+PIDINST` instead of `DOI` in the above examples.

This would be possible, but then (I think) it would make more sense to add an additional field (`_diffrn_component_pid.schema`, for example) rather than writing composite values.

vaitkus: Sorry, my comment was probably not clear. What I am suggesting is that instead of declaring a single data item as a category key (`_diffrn_component_pid.schema`), a composite (DDLm) key consisting of two data items is used instead:
```
loop_
_category_key.name
    '_diffrn_component_pid.framework'
    '_diffrn_component_pid.pid'
```
Think of them as composite table keys in relational databases. This would not require to change any other definitions, but would formally require to explicitly specify not only the pid, but also the framework (with the values of these two items being collectively unique). Would that make sense or would we still need an additional separate 'schema' data item?

cif_core.dic
vaitkus: Thank you for confirming.

cif_core.dic
vaitkus: Thank you for clarifying.

cif_core.dic
```diff
@@ -2488,6 +2488,114 @@ save_diffrn_attenuator.scale
...
+    _description.text
+;
+    The part of the instrumentation to which the PID refers. This
+    value is purely descriptive; the metadata obtained via the PID
+    should be used for rigorous classification and identification.
+;
+    _name.category_id             diffrn_component_pid
+    _name.object_id               details
+    _type.purpose                 Encode
+    _type.source                  Assigned
```
vaitkus:
```suggestion
    _type.purpose                 Describe
    _type.source                  Recorded
```

cif_core.dic
jamesrhester: Making `framework` part of the key is a simple way to enforce it being provided, so I'll update to that.

Changes in the metadata schema I think are outside our scope. Presumably the schema is versioned, and the handle will resolve to an object that provides the schema and version. My intention was to provide just enough information for automated tools to properly access the external metadata via the PID, which I think is achieved?

vaitkus (approved):

PR #603, opened 2026-05-04
Initial draft of released dictionary change summary

jamesrhester:
I've summarised the changes since the last release in the dictionary file `dictionary_audit` loop. The full list is copied in `Detailed_changelog.cif`. Feel free to improve. Once this is merged we will create a release branch.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -29186,4 +29186,5 @@ save_
        equipment.
 
        Added the _dictionary.licensing_SPDX dictionary attribute.
+       Renamed the Head category from 'CIF_CORE' to 'CIF_CORE_HEAD'.
 ;
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -29128,7 +29128,7 @@ save_
 
        Deprecated _publ_contact_author.contact_details, replaced by
        _publ_contact_author.name and _publ_contact_author.address.
-       
+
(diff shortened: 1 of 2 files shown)


cif_core.dic
```diff
@@ -29235,67 +29163,27 @@ save_
...
-
-       Corrected the _name.object_id attribute of _atom_site.fract_symmform
-       and _atom_analytical_mass_loss.temperature_su.
-
-       Moved to ENUMERATION_DEFAULTS for default values.
-
-       Added PIDs for instrumentation.
-
-       Improved DIFFRN_MEASUREMENT category description.
 ;
```
vaitkus, line 29194:
```suggestion
       Renamed the Head category from 'CIF_CORE' to 'CIF_CORE_HEAD'.
;
```

vaitkus: I suggest to maybe also include these two changes as they can be viewed as not-backwards compatible.

vaitkus (approved):
Overall looks good to me.

cif_core.dic
jamesrhester: Because `_atom_type.key` was only invented for dREL and never seems to have been included in any software (e.g. never deposited in COD) I thought it would distract those looking for actionable items. We could leave the head category name change in I suppose.

vaitkus: Understood.

PR #610, opened 2026-06-07
add _enumeration_set.detail values to _space_group.bravais_type

rowlesmr:
_enumeration_set.state for _space_group.bravais_type didn't have any description.

Added _enumeration_set.detail values to _space_group.bravais_type as described in https://journals.iucr.org/a/issues/1985/03/00/a24271/a24271.pdf (de Wolff et al. (1985)), as referenced in ITA.

Removed _description_example.case and .detail as it was just explaining one of the enumeration states.

Didn't update dates, as I didn't change any of the definitions.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11917,20 +11917,20 @@ save_space_group.bravais_type
     loop_
       _enumeration_set.state
       _enumeration_set.detail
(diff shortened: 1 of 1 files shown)


vaitkus (approved):
Looks good to me.

Issue #572, opened 2025-12-09
Harmonise or refine definition of DOIs


nautolycus:
Pull request https://github.com/COMCIFS/cif_core/pull/571 proposes a new item `_audit_conform.dict_DOI` which I hope is not controversial in itself - it simply extends the list of endpoints for locating a dictionary to which the data file conforms, and seems essential if DOIs are to become the preferred way to locate dictionary resources.

I have copied the block of text describing a DOI from other definitions (_e.g._ `save_citation.doi`). These interpret the DOI as an identifier string, and for example give
`    _description_example.case     10.5517/CC6V9DQ`

However, the definition for `save_audit_support.funding_organization_doi` expressly states

>     In accordance with CrossRef guidelines, the full URI of
>     the resolved page describing the funding organization
>     should be given (i.e. including the https://doi.org/
>     component).

As mentioned in Issue https://github.com/COMCIFS/cif_core/issues/275 this appears to be effectively a CrossRef mandate. I think this is unfortunate - a permanent identifier should be recorded independently of a particular resolving infrastructure, and perhaps the dataname should have been named something else, _e.g._ `_audit_support.funding_organization_pURL` (for persistent URL to distinguish it from `_audit_support.funding_organization_URL`).

Anyway, should we

- Leave the other definitions as they stand?
- Change the examples (and explanatory text) in the other definitions to enforce https://doi.org/ prefixes?
- Change the definitions to say that either form is acceptable?

I notice incidentally that the _dictionary.doi string in the current 3.2.0 release version on the IUCr web site says
```
    _dictionary.doi
        https://doi.org/10.1107/cifdic_000002
```

vaitkus:
I would also prefer to have DOIs without the resolver part. I have a few additional questions related to this:
1. If both a DOI and an absolute dictionary URL (`_audit_conform.dict_location`) are provided, but with differing values, which one takes precedence? Of course, the handling of such incompatibilities could be left to the resolving software, however, it might be better to explicitly specify the preference.
2. Is there a need to add a corresponding definition of a dictionary DOI to the `IMPORT` category of the DDLm reference dictionary? This would allow using the dictionary DOI as an identifier in import statements (although a similar result could be achieved by specifying the dictionary location through the DOI resolver).

[...]

PR #612, opened 2026-06-10
Change the usage _audit_support.funding_organization_doi

vaitkus:
This PR was created following the discussion in https://github.com/COMCIFS/cif_core/issues/572.

The updated definition explicitly states that the `_audit_support.funding_organization_doi` data item should record pure DOIs without the resolver part.

Also, I chose to remove the details about URL encoding since this it deals more with how the resolving rather than the representation of the DOI.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.4.0
-    _dictionary.date              2026-05-05
+    _dictionary.date              2026-06-10
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/main/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


cif_core.dic
```diff
@@ -16554,27 +16554,26 @@ save_
...
-    URL-encode certain special characters in the DOI suffix
-    for the URI to resolve correctly through CrossRef
-    (e.g. the number sign '#' must be encoded as '%23').
+    the data collected and analysed in the data block.
+
+    Note that a DOI should be displayed in accordance with
+    the Crossref guidelines (https://www.crossref.org/display-guidelines/),
+    i.e. as a full URL link including the DOI resolver prefix
+    (https://doi.org/). For example, DOI 10.13039/100000064
+    should be displayed as https://doi.org/10.13039/100000064.
```
jamesrhester: Detailed instructions on how to display a DOI don't really belong in a definition, I think. I'd prefer that the second paragraph read "The display form of the DOI may differ from the DOI itself; see, for example, the Crossref guidelines (https://....)."

jamesrhester (changes requested):

cif_core.dic
vaitkus: Updated the text as suggested, please review again.

jamesrhester (approved):

PR #613, opened 2026-06-10
Change all instances of 'CrossRef' to 'Crossref'

vaitkus:
This seems to be the intended capitalisation at least since 2015 (see https://www.crossref.org/blog/the-logo-has-landed/).


rowlesmr (approved):
matches use on Crossref website

PR #616, opened 2026-06-19
Explain use of _diffrn.flux_density

### LINKED ISSUE #611: Harmonise flux definitions with imgCIF
The release candidate includes [a definition for `_diffrn.flux_density`](https://github.com/COMCIFS/cif_core/blob/5abd25b41cae2b080f013861a8fb3dadcc6b7219/cif_core.dic#L389). This shares some definitional intention with a definition in the imgCIF dictionary for [`_diffrn_radiation.beam_flux`](https://mmcif.wwpdb.org/dictionaries/mmcif_img.dic/Items/_diffrn_radiation.beam_flux.html).

Our core definition is for flux density (so per square mm) whereas the imgCIF definition is for flux (integrated over sample). My main concern is that flux and flux density are in different categories.

Our cif_core philosophy of `diffrn_radiation` is that it relates to a source of radiation in general and is not bound to a specific experiment (but of course a specific experiment will have a specific radiation source). If we place `flux density` into `diffrn_radiation`, we are saying that a change in flux density means a new source of radiation. If we place it into `diffrn`, we are saying a change in flux density means a new measurement on the same radiation source.

We currently diverge with mmCIF when thinking about this, as `diffrn_radiation` and `diffrn` are functionally the same category in mmCIF so it doesn't much matter which category it goes into for them.

jamesrhester:
Closes #611.  `_diffrn_radiation.beam_flux` is defined in the imgCIF dictionary and refers to a characteristic of the source, whereas `_diffrn.flux_density` refers to a flux that could change on each measurement, as in ED. Added some words to point users in the direction of `_diffrn_radiation.beam_flux` where they do not need per-measurement values.

Note this is a PR against the release candidate.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2026-06-19
+    _dictionary.date              2026-06-25
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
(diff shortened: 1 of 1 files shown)


vaitkus (commented):
The updated description references `_diffrn_radiation.beam_flux`, but this data item is not currently defined in coreCIF. The item is defined in the mmCIF dictionary, however, this might not be clear to the reader. Maybe be we should redefine it in coreCIF as well?

On the technical side of things, I am not completely sure why this PR is on the rc branch and not the main one. I was under the impression that we were planning to sync all of the non-critical changes from main to the rc just before the release, however, this might have been a misunderstanding from my side.

jamesrhester:
> The updated description references `_diffrn_radiation.beam_flux`, but this data item is not currently defined in coreCIF. The item is defined in the mmCIF dictionary, however, this might not be clear to the reader. Maybe be we should redefine it in coreCIF as well?

Yes, we probably should.

> On the technical side of things, I am not completely sure why this PR is on the rc branch and not the main one. I was under the impression that we were planning to sync all of the non-critical changes from main to the rc just before the release, however, this might have been a misunderstanding from my side.

I did it relative to the RC as, in my mind, this comes under the banner of "issues raised relative to the release candidate" (the only issue), so I was treating it as something in the RC to be fixed, and then we can proceed to release. The difference between main and RC at present is that RC has (in theory) been reviewed by the community at large. Perhaps that's not a distinction worth making, because the present PR is just as much an issue for the main branch, and we could interpret all PRs accepted during the release review period as implicitly addressing issues raised against RC.

jamesrhester:
I have updated the PR to include a definition of `_diffrn_radiation.beam_flux`. It will also require a PR to the Enumeration dictionary to include neutron and electron flux units.

vaitkus (approved):

PR #634, opened 2026-07-19
added enumeration to _atom_type_scat.length_neutron

rowlesmr:
requires https://github.com/COMCIFS/Enumeration_Templates/pull/16
see also https://github.com/COMCIFS/Enumeration_Templates/pull/1

Simple lookup and maintenance, dealing only with element symbol and mass number, not the myriad `_atom_type.symbol` values.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.4.0
-    _dictionary.date              2026-07-10
+    _dictionary.date              2026-07-19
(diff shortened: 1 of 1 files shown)


vaitkus:
Allowing complex values seems like the correct way to go. However, given that we currently interpret Complex numbers as necessarily containing the imaginary component (e.g. 5 + 0i), the type change from `Real` to `Complex` could potentially invalidate some of the older entries (the data name was defined in CIF 1.1). This might actually be an argument in favour of relaxing requirements for Complex numbers.

In the COD, the `_atom_type_scat_length_neutron` is present in around 2500 entries out of over 530 000.

PR #641, opened 2026-08-05
Removed `_diffrn.id` names that belong in multi-block dictionary.

jamesrhester:
Going through core CIF as I revise the core CIF chapter in Vol G, I found that `_diffrn.id` and friends are in core while also being in multiblock. This PR removes them.

This also means removing `_cell.diffrn_id`. My understanding of #442 is that this data name was a mistake and was probably never used. If necessary, we can add it as a deprecated data name in the multi-block dictionary.

Once merged I will backport this to `rc-3.3.0`.


vaitkus:
The [cell-measurement-multi-block.cif](https://github.com/COMCIFS/cif_core/blob/main/examples/cell-measurement-multi-block.cif) file in the examples/ directory should most likely also be moved to cif_core_multiblock repository as it uses these data names. Alternatively, it should be updated to explicitly state that it uses the multiblock dictionary.

I think it somewhat makes sense to at least leave a copy of this example file in this repository since together with [cell-measurement-single-block.cif](https://github.com/COMCIFS/cif_core/blob/main/examples/cell-measurement-single-block.cif) they showcase the old and the new (recommended) way of organizing data side by side.

vaitkus (approved):

PR #648, opened 2026-09-07
Further editorial changes alphabetically as far as AUDIT_SUPPORT.

nautolycus:
Also modified phone and fax definitions to support mobile numbers, and updated examples with valid test numbers.

### DIFF
--- a/cif_core.dic
+++ b/cif_core.dic
@@ -11,7 +11,7 @@ data_CIF_CORE
     _dictionary.title             CIF_CORE
     _dictionary.class             Instance
     _dictionary.version           3.3.0
-    _dictionary.date              2026-09-01
+    _dictionary.date              2026-09-07
     _dictionary.uri
         https://raw.githubusercontent.com/COMCIFS/cif_core/master/cif_core.dic
     _dictionary.ddl_conformance   4.2.0
@@ -15733,7 +15733,7 @@ save_audit.block_doi
     and permits the association of related current data in a
     structured extensible way. A DOI is an implementation
     of the Internet concepts of Uniform Resource Name and
-    Universal Resource Locator managed according to the
+    Uniform Resource Locator managed according to the
     specifications of the International DOI Foundation
     (see https://www.doi.org/).
 ;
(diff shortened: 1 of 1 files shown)


nautolycus:
Some more comment on the phone changes are in order. Ideally I should have prepared this as a separate PR but it evolved while making initially small changes to the definitions. These definitions originally stated

> "The earlier convention of including the international dialing prefix in parentheses is no longer recommended"

but that convention was not reported even in the First Edition of Volume G, so I removed that sentence. But then I thought of mobile phone numbers, that do not have an area code. Investigation suggests that the standard for assigning international telephone numbers is ITU recommendation E.164 (https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-E.164-202602-I!!PDF-E&type=items ). The existing CIF convention deviates from the presentational recommendations in two respects: we do not mandate the use of + to indicate an international dialling prefix, and we do use parentheses. E.164 recommends no punctuation. Since existing legacy CIFs do not have the +, and do have parens, I thought it best not to require adherence to E.164. So I have added the condition "where appropriate" to the reference to area codes. This permits them to be dropped if the user is not sure what constitutes the area code, for example, and especially for mobile numbers where there is not an area code. (There can be network-specific codes, but again a typical user may not know what constitutes that part of a number.)

Provided the international prefix has been used (even without the recommended "+" prefix), any well-formed telephone number conforming to E.164 rules will be globally unique.

The existing dictionary examples (such as 12(34)9477330) are inappropriate because there is no country with dialling code 12, and an attempt to initiate a call to that number may be interpreted incorrectly and end up at a valid subscriber. I have therefore created a number of examples that have valid number representations but are guaranteed never to be allocated to genuine subscribers.

If you are in broad agreement with this approach, please approve and merge this PR. I'll also post this as an issue, and if further refinements are needed (e.g. permitting the "+" as an option), they can be introduced in separate PRs.

Note that I considered the definitional changes, though small, significant enough to change the _definition.update values of these items.

vaitkus (approved):

vaitkus:
The test failure is caused by issue in the check action itself. As such, it should be ignored for now.

nautolycus:
Thank you!

PR #652, opened 2026-09-09
Proposed editorial change to _geom_angle.value definition

nautolycus:


nautolycus:
This one may require some discussion (Issue https://github.com/COMCIFS/cif_core/issues/653), so there is no hurry in merging it.

vaitkus:
I think this PR has been made redundant after merging PR #661 and should therefore be closed.

PR #654, opened 2026-09-09
Editorial tweaks to definitions up to REFINE_LS_CLASS.

nautolycus:


vaitkus (approved):

vaitkus:
Please note that the `PUBL_CONTACT_AUTHOR` is now a looped category, so the descriptions should probably be updated accordingly ("of a contact author" changed to "of the contact author", etc.). Alternatively, the category could be reverted back to a `Set` category (I think it became looped immediately after migration from DDL1 to DDLm). However, I am unsure why the change was made in the first place, maybe to accommodate journals that allow several contact authors?

nautolycus:
> Please note that the `PUBL_CONTACT_AUTHOR `is now a looped category, so the descriptions should probably be updated accordingly

Good catch. I suspect it was changed to Loop to align with `AUDIT_CONTACT_AUTHOR`, where presumably databases allow more than one contact author. I checked with IUCr journals before making the changes in this PR, and they are emphatic that there should be only one article contact author. However, it may be that other publishers are more accommodating (though I foresee administrative difficulties with a policy that allows more than one point of contact).

Anyway, I propose to address this in the next PR by generally changing the introductory sentence to 'The name of _**a**_ contact author...' _etc_. and making similar changes to the `AUDIT_CONTACT_AUTHOR` items. However, I will leave the definite article in place in subsequent references within such a definition, `e.g.` 'This is _**the**_ person contacted by the journal editorial staff.' This may be understood at least by sophisticated English language readers as assigning a (linguistic) categorization to any one of the authors designated as a contact author. The result is a constructive ambiguity that I think permits publishers to have both single- and multiple-agent policies, and preserves the Loop category structure.

PR #656, opened 2026-09-10
Further set of editorial tweaks, to the end of the dictionary.

nautolycus:


vaitkus (approved):

vaitkus:
I think that the "is" in ``Description of the properties of the REFLN reflection list that is not given`` referred to the "description". Nevertheless, changing it to "are" does not alter the meaning significantly.

nautolycus:
> I think that the "is" in Description of the properties of the REFLN reflection list that is not given referred to the "description". Nevertheless, changing it to "are" does not alter the meaning significantly.

Yes, I would parse the original statement as you did, and it would be my personal preference. But increasingly English usage is becoming less rigorous, and the "is" might seem odd to some people and grate upon their ears. I actually made the change to conform to the First Edition, where the decision had been made by the Technical Editor.

It's not a killer, and I want to make another pass through the dictionary to compare against the entire First Edition text. Many technical editing changes in that edition have been reverted (because the DDLm version started effectively from another branch). @jamesrhester and I should review the points of difference and decide whether to implement the technical editor changes from the First Edition, to retain the professional effort that went into intended clarification of the definitions. While I can make such changes to the current dictionary fairly easily (where thought necessary), it might be considered unnecessary effort. Let me take a little while to assess the scale of the possible modifications.

PR #658, opened 2026-09-11
_publ_requested.category updated in line with current practice.

nautolycus:


nautolycus:
If there are no objections, I think this change should also be posted against the main branch.

vaitkus (approved):

vaitkus:
To my best understanding, all changes made in this rc branch will eventually be merged back into the main branch. However, there is probably no harm in issuing a similar PR on the main branch in case you want the updates to appear there faster.

PR #660, opened 2026-09-15
Editorial changes to align with ITG First Edition.

nautolycus:
This is the start of a further pass to align the textual descriptions with those used in the First Edition. This PR covers roughly the first third of the (alphabetically ordered) dictionary.

The descriptions in the current DDLm dictionary derive from the draft dictionary originally presented in the COMCIFS 2013, but that took as its starting point a set of definitions pre-dating the publication of the First Edition of Volume G. This pass will superimpose the technical editing changes made for publication, where appropriate.


cif_core.dic
```diff
@@ -10279,7 +10292,8 @@ save_chemical_conn_bond.type
     _definition.update            2012-11-22
     _description.text
 ;
-    Code for the chemical bond type.
+    The chemical bond type associated with the connection between the two sites
+    chemical_conn_bond.atom_1 and *_2.
```
vaitkus:
```suggestion
    _chemical_conn_bond.atom_1 and *_2.
```

cif_core.dic
```diff
@@ -15835,7 +15851,7 @@ save_audit.creation_date
     _definition.update            2019-03-26
     _description.text
 ;
-    The timestamp of the data revision.
+    The date on which the data block was created.
```
vaitkus, line 15854: The word `date` might not be the best fit here as the content type now allows either a date or a full timestamp (see examples).

cif_core.dic
```diff
@@ -17331,7 +17355,8 @@ save_citation.id
     _definition.update            2025-07-31
     _description.text
 ;
-    Unique identifier to the CITATION list. A value of 'primary'
+    The value of citation_id must uniquely identify a record in
```
vaitkus:
```suggestion
    The value of _citation.id must uniquely identify a record in
```

cif_core.dic
```diff
@@ -21967,8 +22002,10 @@ save_ATOM_SITE
     _definition.update            2023-02-03
     _description.text
 ;
-    The category of data items used to describe atom site information
-    used in crystallographic structure studies.
+    Data items in the ATOM_SITE category record details about
+    the atom sites in a crystal structure, such as the positional
+    coordinates, atomic displacement parameters, and magnetic moments
```
vaitkus, line 22007: Do `magnetic moments and directions` here refer to the extensions provided in the `magCIF` dictionary? Wouldn't it be a bit confusing to a person reading a standalone version of `coreCIF` to encounter references to magnetic data even though the definitions themselves are not provided directly in the dictionary?

cif_core.dic
```diff
@@ -24432,7 +24475,8 @@ save_atom_sites.special_details
     _definition.update            2012-11-20
     _description.text
 ;
-    Information about atomic coordinates not coded elsewhere in the CIF.
+    Additional information about the atomic coordinates not coded elsewhere
+    in the CIF.
```
vaitkus, line 24479: To my best understanding, DDLm is now considered a format-agnostic, so maybe the reference to CIF could be dropped/repleaced here?

cif_core.dic
```diff
@@ -26330,8 +26375,9 @@ save_atom_type_scat.dispersion_real
     _definition.update            2012-11-20
     _description.text
 ;
-    The real component of the anomalous dispersion scattering factors
-    for this atom type and radiation by _diffrn_radiation_wavelength.value
+    The real component of the anomalous-dispersion scattering factor
+    for this atom type and the radiation specified by
+    _diffrn_radiation_wavelength.value
```
vaitkus, line 26380:
```suggestion
    _diffrn_radiation_wavelength.value .
```

cif_core.dic
```diff
@@ -10138,7 +10143,9 @@ save_chemical_conn_atom.type_symbol
     _definition.update            2021-10-27
     _description.text
 ;
-    A code identifying the atom type.
+    A code identifying the atom type. This code must match an
+    _atom_type.symbol code in the ATOM_TYPE list or be a
+    recognizable element symbol.
```
vaitkus, line 10148: I am unsure if the `or be a recognizable element symbol` part is compatible with the formal definition. That is, since the definition explicitly links to `_atom_type.symbol` using the `_name.linked_item_id` attribute, general purpose validators would complain about recognizable element symbols that are not provided in that list.

cif_core.dic
```diff
@@ -10490,9 +10504,11 @@ save_chemical_formula.sum
     _definition.update            2012-11-22
     _description.text
 ;
-    Chemical formulae in which all discrete bonded residues and ions are
+    See the CHEMICAL_FORMULA category description for the rules for writing
```
vaitkus, line 10507: This reads to me as if the `CHEMICAL_FORMULA` description contains specific rules for writing `chemical formulae in which all discrete bonded <...>` while it practice it seems to contain the general rules for writing formulae. Maybe this could be better written in a separate sentence after the main definition. Something similar is done in the `_chemical_formula.moiety` definition (`See above CHEMICAL_FORMULA for rules for writing chemical formulae.`)

vaitkus (commented):
I suggested a few minor changes and raised a few notes. The notes do not necessarily require full resolution here, they can be spun off into separate issue if needed.

If my review alone is sufficient here, I would be happy to merge the PR once ready (e.g. after the minor changes).

cif_core.dic
jamesrhester: I agree that it would be best not to mention CIF syntax explicitly.

cif_core.dic
nautolycus: Agreed. How about 'The date on which the datablock was created. If necessary, a creation time may also be specified, to an appropriate precision." The point being that a date will usually suffice, but more granular timestamping will depend on the purpose for which each data block was generated (e.g. in serial crystallography).

cif_core.dic
nautolycus: Agreed. I suspect the original wording was designed to encourage the addition of other properties, whether through new core anmes or (at that time unknown) extension dictionaries. The use of "such as" allows truncation of the sentence as "such as the positional coordinates and atomic displacement parameters."

cif_core.dic
nautolycus: OK. "... not recorded elsewhere"? "Recorded" is a better word than "coded" here, anyway.

cif_core.dic
nautolycus: I think the convention is to close up the final full stop even after a data name, but I'll check here and elsewhere for the next pass.

cif_core.dic
nautolycus: I think the (potential) issue is that the atom site may be characterized by an extended `_atom_type.symbol` value such as Ca2+, but for the purposes of a chemical diagram an author might wish to simply use Ca. But formally that can be addressed by an entry for Ca in the `ATOM_TYPE` list, if there is not already one, coupled with an explanation, if necessary, in `_atom_type.description` that it used only as a label for a chemical diagram. So I'm OK with dropping the "or be..." phrase.

cif_core.dic
nautolycus: OK. I'll review this and the other `_chemical_formula.*` terms with this in mind.

nautolycus:
@vaitkus : Thanks for your vigilance and good sense on these. As this is still work in progress, please accept the current PR if you're reasonably happy with my proposals, and I'll implement them in the next iteration.

vaitkus (approved):

cif_core.dic
vaitkus: This was addressed in PR #661.

cif_core.dic
vaitkus: Issue was resolved in PR #661.

cif_core.dic
vaitkus: Issue was resolved in PR #661.

cif_core.dic
vaitkus: Comment was addressed in PR #661.

The convention was indeed to close up the final full stop.

cif_core.dic
vaitkus: Comment was addressed in PR #661.

Sorry, I somehow missed your response, the updated text looks great.

cif_core.dic
vaitkus: PR #661 modified several definitions relating to chemical formulae, but seemingly not this specific definition.

nautolycus: Sorry, I seem to have missed this one. Proposal for the next PR:

```
    Formula in which all discrete bonded residues and ions are
    summed over the constituent elements, following the ordering given
    in general rule (5) of the CHEMICAL_FORMULA category description.
    Parentheses are not normally used.
```

vaitkus: The comment was addressed as described in PR #667. Thanks!

PR #661, opened 2026-09-16
More First-Edition-compatible editorial changes.

nautolycus:


nautolycus:
These cover roughly the middle third of the dictionary.

cif_core.dic
```diff
@@ -8992,9 +9054,9 @@ save_EXPTL
     _definition.update            2024-05-15
     _description.text
 ;
-    The category of data items used to specify the experimental work
-    prior to diffraction measurements. These include crystallization
-    crystal measurements and absorption-correction techniques used.
+    Data items in the \cifcategory{EXPTL} category record details about the
```
vaitkus, line 9057:
```suggestion
    Data items in the EXPTL category record details about the
```

cif_core.dic
```diff
@@ -4074,6 +4088,15 @@ save_DIFFRN_REFLN
 ;
     The category of data items which specify the reflection measurements,
     prior to data reduction and merging.
```
vaitkus, line 4090: These lines should have probably been removed, i.e. replaced by the newly added text below.

cif_core.dic
```diff
@@ -4393,7 +4424,8 @@ save_diffrn_refln.elapsed_time
     _definition.update            2012-11-26
     _description.text
 ;
-    Elapsed time from the start to the end of the intensity measurement.
+    Elapsed time from the start of the diffraction experiment to the
+    measurement of this intensity.
```
vaitkus, line 4428: The "to the measurement" seems a bit ambiguous whereas the previous definition specifically identified the moment in time as the end of the measurement. Of course, that does not make much of a difference if the measurement is extremely short.

cif_core.dic
```diff
@@ -13641,8 +13712,9 @@ save_geom_angle.value
     _definition.update            2020-03-13
     _description.text
 ;
-    Angle defined by the atoms located at atom_site_x/site_symmetry_x for
-    x = 1,2,3. The vertex atom is at site x = 2.
+    The value of the angle defined by the listed sites.
+    The vertex atom is identified by _geom_angle.atom_site_label_2
+    (and accompanying symmetry operation, if relevant).
```
vaitkus, line 13717: I would say the the symmetry operation is always used when identifying a site in the geometry categories, either explicitly (given in CIF) or implicitly (not given in CIF thus defaulting to 1_555). Therefore, it might be better to lose the parenthesis and the "if relevant" part, i.e.:
```
... _geom_angle.atom_site_label_2 and the accompanying symmetry operation code.
```
Also, I think that the term `symmetry operation code` should be used here instead of `symmetry operation` since these refer to different types of transformations (i.e. 2_565 vs -x,-y,-z). The term symmetry operation code is already used to refer to such values in the site symmetry item definitions (see site_symmetry save frame from the templ_attr.cif).

vaitkus (approved):
A few minor issues and comments, looks good overall.

nautolycus:
Great! I agree with all of these and shall incorporate into my next PR. Many thanks.

vaitkus:
I would prefer for the straightforward changes (e.g. removal of duplicated text) to be applied directly in this PR as it is a bit easier to track, however, I have no objections to merging it as is.

cif_core.dic
vaitkus: Comment addressed in PR #663.

cif_core.dic
vaitkus: Comment addressed in PR #663.

cif_core.dic
vaitkus: Comment addressed in PR #663.

cif_core.dic
vaitkus: Comment addressed in PR #663.

PR #662, opened 2026-09-16
Fix drel code

vaitkus:


cif_core.dic
```diff
@@ -29768,4 +29768,7 @@ save_
        updated examples with valid test numbers.
 
        Updated _publ_requested.category to lates IUCr journal requirements.
+
```
jamesrhester: I would not add anything to the summary information. This summary has been condensed from the full summary and should only contain significant changes.

jamesrhester (commented):
Amazing that this error has been in the dictionary for so long.

cif_core.dic
vaitkus: Ok, I have removed the changelog sentence.

jamesrhester (approved):

PR #666, opened 2026-09-18
Fix minor formatting issues

vaitkus:
I think that some of these were accidentally introduced during the few latest rounds of editorial revisions.

@nautolycus Also, I noticed that in the bibliographic references, the page numbers are sometimes separated by a regular dash ('-') and sometimes by a long dash ('–'). I did not attempt to address this in this PR, since I am not sure which symbol is the preferred one.


nautolycus:
Thank you @vaitkus . Yes, some of the infelicities arise because I am back-porting changes from the First Edition that were actually easiest for me to locate using the formatted (LaTeX) dictionary.

Given that we are happy to introduce Unicode in this version, it would be good to settle on the long dash, but I consider that a second-order style convention that can be fixed on an 'as-noticed' basis. Either choice will be handled correctly in my typesetting procedures. I do commend you on your sharp eyesight!

vaitkus:
Understood.

